PR #1566, opened 2015-05-03
swig tool can handle -outdir parameter

yag00:
Swig tool can easily handle swig -outdir parameter with this patch

### DIFF
--- a/playground/swig/wscript
+++ b/playground/swig/wscript
@@ -92,10 +92,10 @@ def build(bld):
 
 	from waflib.extras import swig
 
-	srcdir = bld.path.get_bld().make_node('extend/java/hmm') # destination for generated java file (without the packages!)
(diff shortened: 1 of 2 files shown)


ita1024:
Why is this necessary once again?

yag00:
Not necessary but it's easier to generate file where you want them instead of doing some copy after generation.

ita1024:
There was a TODO note concerning this. How do you use it (it would be best if we had a testcase in playground/swig), and on which operating systems have you used it?

yag00:
I tested it on Ubuntu 14.04 and Fedora 20 64bit. I just commited the updated playground/swig test case in my master branch.  Work with both block case...

ita1024:
We will have to try on windows systems.

Issue #1568, opened 2015-05-05
Hook to suppress manifest generation when building cpp-clr 


JodyGoldberg:
Msvc supports a C++ variant cpp-cli via the flag '/clr'.   The associated libraries are built in mostly the same fashion as C++ with two exceptions
1) They support a '/FU<clr_dll>' argument the force the usage of a particular managed library
2) They do not have manifest files

After some playing it seemed cleanest to add a hook to ccroot:use_rec to capture that the target is a CPPCLI lib and a second hook in ccroot:process_use to use that hook to suppress the manifest.

Any suggestions on a better approach?

ita1024:
This looks like an os-specific optimization.
1. I do not understand how doing the above suppresses the creation of a manifest file. It seems that the only thing done is to add object files instead of linking the library
2. If the library is given as argument though something else will occur - x in names - and i am not sure that this is desirable
3. The functionality adds complexity and is guaranteed to break in the future
4. The same results could be achieved easily through using an intermediate "objects" target

ita1024:
...

Issue #1567, opened 2015-05-05
Argument ConsoleWrite in LPDWORD not LPINT


JodyGoldberg:
Causes exception when using ipdb on windows with 64bit python

ita1024:
Thanks a4c2370ee4669ca7cce30768dd8211295a4eda78

PR #1569, opened 2015-05-05
Cppcli

JodyGoldberg:
Patches for #1567 and #1568


ita1024:
The LPWORD change is good, but the other is not. It would be best if two separate pull requests had been made :-( I have merged the change manually.

PR #1570, opened 2015-05-06
cpplint: fixed import.

srouquette:
cpplint: fixed import. The version on pypi is now up to date and works with Python3.


ita1024:
The original code used to work, it might be safer to try one import and then the other.

PR #1572, opened 2015-05-08
build package: don't interpret \\ when joining the string, let Popen do that.

srouquette:
On Windows, when building waf with --make-waf, an error is produced.
It comes from this change: 588f809ffa4dd514ea90bdcd0341d9baf508784f
PREFIX on Windows will return something like "C:\\\\users\\\\..."
\\\\ was resolved with the join, and Popen would try to interpret \uXXXX as a unicode character.


ita1024:
I think this change does not really fix the problem. Can you please try 802e937d7328a0ff87516147d1f80dceb982120b ?

srouquette:
yes, you're fix work. didn't know %r  :)

ita1024:
This will also work with paths containing apostrophes.

Issue #1574, opened 2015-05-14
duplicated LINKFLAGS if both C and C++ compilers are configured


che2:
The contents of LINKFLAGS or LDFLAGS environment variables are duplicated if both a c compiler and a c++ compiler are configured using:

    def configure(ctx):
        ctx.load('compiler_c')
        ctx.load('compiler_cxx')

The reason is that both of the c and c++ compiler plugins (say gcc.py and gxx.py) calls the function conf.link_add_flags(), which is defined in c_config.py:

    def link_add_flags(conf):
        """
        Add LINKFLAGS / LDFLAGS from os.environ to conf.env
	"""
        conf.add_os_flags('LINKFLAGS')
        conf.add_os_flags('LDFLAGS')

This results in LINKFLAGS being added twice to ctx.env.

    $ LINKFLAGS='-lblah' waf configure
    $ cat build/c4che/_cache.py
      ...
      LINKFLAGS = ['-lblah', '-lblah']
      ...

ita1024:
Does this actually pose a problem?

che2:
This is not a deal breaker,  but it is confusing and obtrusive to debugging by unnecessarily increasing the length of the link line.  I think changing the behavior so that the LINKFLAGS is no longer duplicated makes a lot of sense.

che2:
I have attempted to create a fix at #1575.  The fix simply prevents flags from being imported more than once. I think this should be safe, as I can't think of a reason why you would want to ever duplicate an environment flag.

ita1024:
That pull request introduced a new problem and changed the meaning of "add_os_flags". Adding a new parameter is necessary, try f048d6d20f3b274e65d397ee87b25927346f0ff5

che2:
Confirmed that commit f048d6d works. Thanks.

PR #1575, opened 2015-05-15
Fix duplicate flags from environment #1574

che2:
Build flags from the os environment might be duplicated when imported to
waf.  For example, LINKFLAGS will be defined twice if C and C++
compilers are both configured.  This is behavior is confusing and may
cause problems during debugging.

#1574


ita1024:
This will not work when configuring variants ("self.env" can point to several objects).

PR #1577, opened 2015-05-26
Fix the '--threading' option for valac.

arteymix:
The actual compiler option is named `--thread`, not `--threading`.

Renames the waf option for 'thread': it shouldn't be breaking anyone as it was not even working.

Maybe `--threading` this was an old option for valac? It would be really nice to double check before merging.


ita1024:
I looked in previous valac versions. This has never worked and has never been tested.

arteymix:
@ita1024 Thank you :+1:

PR #1578, opened 2015-05-26
Avoiding attribute errors for Fortran modules

haraldkl:
Occasionally I end up with missing signature attribute errors in my Fortran projects after a failed compilation. The behavior is a little weird, and I can not provoke it with the Fortran example in the demos.
Here is the backtrace of the occuring error:

Traceback (most recent call last):
  File "/Users/klimach/Projects/apes/seeder/.waf-1.8.7-315a7b482a73bed5780df933adccfde1/waflib/Runner.py", line 150, in task_status
    return tsk.runnable_status()
  File "/Users/klimach/Projects/apes/seeder/.waf-1.8.7-315a7b482a73bed5780df933adccfde1/waflib/Tools/fc.py", line 46, in runnable_status
    ret=tsk.runnable_status()
  File "/Users/klimach/Projects/apes/seeder/.waf-1.8.7-315a7b482a73bed5780df933adccfde1/waflib/Tools/fc.py", line 40, in runnable_status
    return super(fc,self).runnable_status()
  File "/Users/klimach/Projects/apes/seeder/.waf-1.8.7-315a7b482a73bed5780df933adccfde1/waflib/Task.py", line 313, in runnable_status
    new_sig=self.signature()
  File "/Users/klimach/Projects/apes/seeder/.waf-1.8.7-315a7b482a73bed5780df933adccfde1/waflib/Task.py", line 302, in signature
    self.sig_implicit_deps()
  File "/Users/klimach/Projects/apes/seeder/.waf-1.8.7-315a7b482a73bed5780df933adccfde1/waflib/Task.py", line 386, in sig_implicit_deps
    if prev==self.compute_sig_implicit_deps():
  File "/Users/klimach/Projects/apes/seeder/.waf-1.8.7-315a7b482a73bed5780df933adccfde1/waflib/Task.py", line 424, in compute_sig_implicit_deps
    upd(k.get_bld_sig())
  File "/Users/klimach/Projects/apes/seeder/.waf-1.8.7-315a7b482a73bed5780df933adccfde1/waflib/Node.py", line 471, in get_bld_sig
    self.cache_sig=ret=self.sig
AttributeError: sig

Somehow, after (some) failed compilations, there seems to be a "prev" instance of the task available, but the *.mod file is missing and the corresponding node does not have a "sig" attribute. I am not sure, wether this should be fixed beforehand that this can never happen.
But catching the missing attributes here and causing a rescan of the dependencies seems to work fine.

I guess, this is related to #1546.


ita1024:
I am not convinced. Why does this happen?

haraldkl:
I don't know. I first thought it might be a concurrency problem, but it also happens with -j1. It might be due to some pre-processing I am doing: https://bitbucket.org/apesteam/treelm/src/6c00d235a1a5857f34a8e4d34b1ee0467cc5ee0d/external/coco.py?at=default, but it also happens for files without pre-processing. I'll need to dig further into it. However, it looks like there need to be at least 2 subsequent failed compilations, to provoke this behavior.

haraldkl:
In are_implicit_nodes_ready the dependencies are checked against the outputs of all tasks. However, dependencies are on the module files *.mod, while the outputs in the list of tasks seem to be only the object files. Thus, none of the dependencies is found in the mapping of outputs to tasks (bld.dct_implicit_nodes) and it is assumed, that all the tasks where already done.

This is the same, in the working case (after a ./waf clean) and in the broken case. The difference is, that in a clean build, the run_after of the tasks is properly set before entering the are_implicit_nodes_ready routine, while in the broken case those are empty before.
The strange thin to me, is that the tasks once (in the run_after) show all the outputs and once in the bld.dct_implicit_nodes within the same waf run.

ita1024:
Provide a testcase. This change affects many other areas and does not fix the issue, so it will not be merged.

PR #1581, opened 2015-05-29
Fix a bug for Vala when two tasks share build files.

arteymix:
If two tasks, let's say a shared and a static library, share the same
sources `*.vala`, they will generate the same `*.c` files, but since it's
part of the valac task to move these files, they won't be available for
the second task.

This is a naive fix that simply replaces `shutil.move` for `shutil.copy`.
A good fix would probably generate the files at the right locations and
simply avoiding the move call.

### DIFF
--- a/waflib/Tools/vala.py
+++ b/waflib/Tools/vala.py
@@ -32,7 +32,7 @@ class valac(Task.Task):
 
 		for x in self.outputs:
 			if id(x.parent) != id(self.outputs[0].parent):
-				shutil.move(self.outputs[0].parent.abspath() + os.sep + x.name, x.abspath())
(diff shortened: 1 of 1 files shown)


ita1024:
There is a chance that a truncated file will be copied (parallel builds). Is there any way to have another file written?

arteymix:
@ita1024 using `--directory`, we can specify where the build files are generated and thus, avoid moving them, the dependency handling should do the rest

I really wish I would understand waf source better.. I'll try to dig a fix, but if it's obvious, let me know.

ita1024:
This PR introduces a race condition. We cannot merge this (discussed in #1601 ).

PR #1585, opened 2015-06-14
[macplist]: interpolate file

caitp:
Rather than using an inline interpolated heredoc string, or generating
the file in a pre-build step, allow the macplist task itself to
perform string interpolation on the contents of a plist file.


caitp:
Hey, it's been a while since I've gotten any code upstream in waf --- It would be nice to get this in, so that I can get the inline string formatting out of my wscripts =)

PR #1586, opened 2015-06-15
[macapp] support ant-globbing mac_resources + deep resource directories

caitp:
Previously, resource sources were always installed directly under the 'Resources'
directory. It is now possible to specify a root directory for resources, in order
to install deep directory structures, suitable for localization directories or
other resource directories.

### DIFF
--- a/tests/mac_resources/src/Resources/MainMenu.nib
+++ b/tests/mac_resources/src/Resources/MainMenu.nib
@@ -1,0 +1,1 @@
+fake MainMenu.nib
--- a/tests/mac_resources/src/Resources/controls/Carousel.nib
+++ b/tests/mac_resources/src/Resources/controls/Carousel.nib
@@ -1,0 +1,1 @@
+fake Carousel.nib
--- a/tests/mac_resources/src/Resources/controls/Poster.nib
+++ b/tests/mac_resources/src/Resources/controls/Poster.nib
@@ -1,0 +1,1 @@
+fake Poster.nib
--- a/tests/mac_resources/src/main.c
+++ b/tests/mac_resources/src/main.c
@@ -1,0 +1,3 @@
+int main(int argc, char** argv) {
+	return 0;
+}
--- a/tests/mac_resources/wscript
+++ b/tests/mac_resources/wscript
@@ -1,0 +1,111 @@
+#! /usr/bin/env python3
+
+import platform, sys, inspect
+
+top = "."
+out = "bin"
+
+def options(opt):
+	opt.load("compiler_c")
+
+def configure(conf):
+	conf.load("compiler_c")
+
+def build(bld):
+	# Only testing feature on Darwin
+	if (platform.system() != "Darwin"):
+		return
+
+	bld.program(
+		features="c cprogram",
+		target="SpaceSeparatedListTest",
+		source="src/main.c",
+		mac_app=True,
+		mac_resources="""
+			src/Resources/MainMenu.nib
+			src/Resources/controls/Carousel.nib
+			src/Resources/controls/Poster.nib
+			""")
+	bld.add_post_fun(test1)
+
+	bld.program(
+		features="c cprogram",
+		target="AntGlobStarStarTest",
+		source="src/main.c",
+		mac_app=True,
+		mac_resources=bld.path.ant_glob("src/Resources/**"))
+	bld.add_post_fun(test2)
+
+	bld.program(
+		features="c cprogram",
+		target="AntGlobRelativeToResourcesPathTest",
+		source="src/main.c",
+		mac_app=True,
+		mac_resources_path="src/Resources",
+		mac_resources=bld.path.ant_glob("src/Resources/controls/**", dirs=True))
+	bld.add_post_fun(test3)
+
+	bld.program(
+		features="c cprogram",
+		target="AntGlobExcludeTest",
+		source="src/main.c",
+		mac_app=True,
+		mac_resources=bld.path.ant_glob("src/Resources/**", excl="**/Carousel*"))
+	bld.add_post_fun(test4)
+
+	bld.program(
+		features="c cprogram",
+		target="LegacyDirectoryTest",
+		source="src/main.c",
+		mac_app=True,
+		mac_resources="src/Resources")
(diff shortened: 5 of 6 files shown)


ita1024:
The line 105 looks strange:
  nodes = [x for x in node.ant_glob('**') if x.abspath() in resource_paths]
1. If "resource_paths" is not provided, should it not default to providing all files by default as the previous implementation?
2. What is the point of calling ant_glob if all the files are already known in advance? is the intention to check whether the nodes found "start with" / "end with" one of the "resource_paths" entries?

caitp:
> 1. If "resource_paths" is not provided, should it not default to providing all files by default as the previous implementation?

If `resource_paths` is (perhaps confusingly) distinct from `mac_resources_path` --- It's just a way of making sure that only paths matched by the glob are processed, if a glob is used. It might benefit from an extra test-case just to make sure it works right when the `nodes = ant_glob('**')` path is taken

> 2. What is the point of calling ant_glob if all the files are already known in advance? is the intention to check whether the nodes found "start with" / "end with" one of the "resource_paths" entries?

I assume you mean globbing the subdirectory --- since previously globs weren't supported, it wasn't possible to know the files in advance, I think this is just there for legacy compat. Might not be needed anymore, thoughts?

ita1024:
Ah, so there is "resources_path" besides "resource_paths" now. I still do not understand what the code is doing, if a glob is used then why filter it with "if x.abspath() in resource_paths"? does this not disable the glob?

caitp:
Suppose I have a glob like `ant_glob("src/Resources/**", exclude="src/Resources/Foo/**"`, right --- The glob shouldn't contain any paths within "src/Resources/Foo", but the secondary glob (used to not break the api that already existed) will still include the path. So it's just a check to make sure the path is not intended to be excluded

ita1024:
If a folder is given as "mac_resources", then the glob will always only return the folder itself? I do not have a mac to try the code in question and it is not easy to understand.

caitp:
I'll add a 4th test case that should hopefully clarify it, hang on

ita1024:
I cannot execute the testcases: no mac here.

caitp:
Yeah, but you should be able to read the code and see what I'm talking about, I think

caitp:
Test is pushed --- so, with the legacy non-glob api, the `Carousel.nib` would be written into the Resources directory. I don't want to break the old behaviour in case people depend on it, and there's no way to know (as far as I can see) whether an ant_glob or something else was used. So it does the secondary glob, and makes sure that if a proper glob was used, it doesn't include anything that was otherwise excluded

caitp:
just fyi, I'm working on a patch to get tests running on CI so that it's a bit easier to verify this stuff --- Travis-CI seems to be painful for multi-arch testing right now, but mac-only should be able to test most of what is needed

ita1024:
We already have a few CI servers running. Mac OSX only runs on mac hardware due to licensing reasons.

ita1024:
Regarding the tests, using mac_resources="src/Resources/" used to glob all files recursively under that folder to include them. This will not work anymore then?

caitp:
No, that still works, shall I add one more test case? (I might suggest landing the Travis-CI integration thing before this just so that you can more easily verify it)

ita1024:
Does demos/mac_app still work after these changes?

You could actually remove the limitations of your tests so that they run on Linux too, I am not convinced that Travis is actually required.

caitp:
something in the test I added is broken so I'll have to look at that first, but I am 90% sure it's just a bug in the test and not implementation. I'll take a look in a few minutes

caitp:
> Does demos/mac_app still work after these changes?

Behaviour of the demo is verified by the first test case, it still works. The `mac_resources="someDir"` thing was broken, but I have a patch in progress, and it now exactly matches the behaviour that is upstream

ita1024:
The demo and the first test case differ quite a bit. Have you actually built demos/mac_app and verified that the files are included?

caitp:
yes, and the files are included

ita1024:
I now understand that you have rewritten the patch. Although the code has become more readable, major problems remain:

1. Recycling the "mac_resources" attribute for other purposes is a questionable idea. Since more than one "mac_resources_path" might be needed by node, it would be more adequate to consider a list of tuples (path, parent) for example
2. The tests will only run on Mac platforms, even though the code could be made portable with little effort. I understand that this is to push your Travis pull request, but this is unnecessary.
3. Code containing bare "except:" clauses will not be merged. Catching Keyboard interruptions and memory errors in such context is not acceptable.

caitp:
It has nothing to do with pushing the Travis patch, but this code is hosted on github now, and Travis is basically the de facto ci service which integrates easily with github, is free, and has Mac vms available, all of which are good things. If Jenkins or whatever other services are being used, it's a good thing to set up service hooks and make them visible.

I'll fix up the bare except. Re: the multiple mac_resource_paths, certainly it's possible to need more than one, or write globs  that produce weird  relative paths. I think for now, to keep it simple, I would opt to just leave it and iterate on that, because it increases the complexity to handle those cases. But if you think those are valuable, and block landing it, I'll fix that too

ita1024:
I am not insisting on using tuples, but not re-using mac_resource_paths is necessary. We can then change/merge the behaviour of the old attribute in waf 1.9 but not in the middle of the 1.8 series.

Can you perform the changes (new attribute, proper exception handling, portable test), or do you want me to do them?

caitp:
I think I can make the change if I understand what you're asking for re "new attribute".

If I understand you, you're saying "leave mac_resources alone and create a new glob-aware attribute that can be used along-side or in place of mac_resources", is that right?

ita1024:
Yes.

ita1024:
It looks like I have to do it myself... The revision 933feab352f11179e63b17f762648d307d26e68d should provide a good start.

In future submissions, try to provide code that does not seem to be inspired from the [unmaintainable code guide](http://mindprod.com/jgloss/unmain.html):
* obfuscated variable names (1-letter difference)
* recycling attributes with a different purpose
* meaningless comments
* catching system exceptions

ita1024:
The changes are in Waf 1.8.12.

PR #1587, opened 2015-06-16
[ci] add test-runner for travis-CI

caitp:
This runs the test suite on a mac, and also provides a helpful script for
running the entire suite of tests locally. A code for gray ANSI colors
is also included in the patch, because it looks nice on the Travis UI to
indicate skipped tests

For a passing build, see https://travis-ci.org/caitp/waf/builds/66960282 (it took a lot of tries to get the configuration working correctly, and some of these tests might still be a bit flakey)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,11 @@
+language: objective-c
+os:
+  #TODO: Travis-CI is broken for any kind of multi-platform testing at the moment :(
+  # - linux
+  - osx
+
+before_script:
+  - source ./scripts/ci/travis.sh
+script:
+  - ./waf-light --make-waf
+  - ./tests/run_tests.py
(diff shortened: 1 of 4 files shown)


caitp:
yeah there are flakes, which could probably be improved --- still it's nice to actually test on a mac =)

caitp:
This now retries failures a few times by default (10 at the moment, but 2-3 is probably enough), https://travis-ci.org/caitp/waf/builds/67057861 --- and it keeps the build green

ita1024:
The previous pull request illustrates the problem: while the tests seem to work, the environment does not allow for control over the environment and the virtual machine (it is just someone else's computer).

What the project really needs here is a mac mini part of our current infrastructure. Merging this pull request is therefore not going to happen.

PR #1591, opened 2015-06-28
Fix exec_command and cmd_and_log when cmd is tuple

yotann:
Before, they would crash because Python was trying to format `%r` using the tuple, which had more than one element.


ita1024:
How did you end up with tuples in those commands? They are really meant to accept either list or string types.

yotann:
I called it directly with a tuple, assuming a tuple would work the same way as a list.

ita1024:
Oh dear. Avoid tuples unless you actually mean immutable data.

PR #1593, opened 2015-07-01
ConfigSet propsed new methods

daniele-niero:
I'd like to propose fw new methods for ConfigSet.

The first two are related to the ConfigSet's dict interface. They are __iter__ and iteritems.
They do exactly what their counterparts in a standard dict do.

```python
def configure(conf):
        for k in common_env:
            print k

        for k, v, in common_env.iteritems():
            print k, v
```

the last method allows a deep copy of a ConfigSet object.

consider this configure:
I'd like to use conf.env as a shared, common environment and change only the relevant parts in any subsequent variants.

```python
def configure(conf):
        conf.env["AAAA"] = "Whatever"
        conf.load('compiler_c')
        conf.env.CFLAGS = []

        conf.setenv('release', conf.env.derive())
        conf.env.CFLAGS += ['-O2']

        conf.setenv('debug', conf.env.derive())
        conf.env.CFLAGS += ['-g']

        conf.setenv('special', conf.env.derive())
        conf.env.CFLAGS.append("Special")
```

Unfortunately I will end up with the same 3 environments here, because derive() creates only a shallow copy. In fact in this example derive() is completely useless as I'm changing yes a different ConfigSet, but all of them shares the same objects.

consider this instead:

```python
def configure(conf):
    conf.env["AAAA"] = "Whatever"
    conf.load('compiler_c')
    conf.env.CFLAGS = []

    common_env = conf.env.deepcopy()
    
    conf.setenv('release', common_env.deepcopy())
    conf.env.CFLAGS += ['-O2']
    
    conf.setenv('debug', common_env.deepcopy())
    conf.env.CFLAGS += ['-g']
    
    conf.setenv('special', common_env.deepcopy())
    conf.env.CFLAGS.append("Special")
 ```

this will allow me to load a tool, compiler_c in this example, only once, then settings some common environments variable.
Later I can copy the common environment for each new variant being sure that they will not interfere with each other environment.

### DIFF
--- a/waflib/ConfigSet.py
+++ b/waflib/ConfigSet.py
@@ -131,6 +131,14 @@ class ConfigSet(object):
 		else:
 			del self[name]
 
+	def __iter__(self):
+		merged_table=self.get_merged_dict()
+		return iter(merged_table)
+
+	def iteritems(self):
+		merged_table=self.get_merged_dict()
+		return merged_table.iteritems()
+
 	def derive(self):
 		"""
 		Returns a new ConfigSet deriving from self. The copy returned
--- a/waflib/ConfigSet.py
+++ b/waflib/ConfigSet.py
@@ -340,3 +340,15 @@ class ConfigSet(object):
 		"""
 		self.table = self.undo_stack.pop(-1)
 
+	def deepcopy(self):
+		"""
+		Create a deep copy (as opposed to .derive) of this ConfigSet.
+		"""
+		merged_table=self.get_merged_dict()
+
+		other = ConfigSet()
+		for key, value in merged_table.iteritems():
+			if key!='undo_stack':
+				other[key] = copy.deepcopy(value)
+		return other
+

### CODE REVIEW

**Comment by ita1024** [comment]
Your examples do not reflect the recommended usage: this data structure is meant for shallow copies and one should use the methods append_value, append_unique and prepend_value. The methods that you suggest can be added, but they will not be advertised for the reason described.

Now there are also a few details to sort out:
 * The deep copy can be achieved easily by calling derive() and detach()
 * The __iter\_\_ method contains an unnecessary temporary variable - just return iter(self.get_merged_dict()) directly?
 * The iteritems method does not exist in Python3, so I do not want this one

**Comment by daniele-niero** [comment]
Sounds all reasonable observations to me.

However with append_value, append_unique and prepend_value I couldn't
achieve the same thing as deep coping the all object. Same things for
derive detach.
Maybe I missed something obvious, I wouldn't be surprise as it is normally
easy to overlook some features when learning a new library.

Maybe an example that shows how to do it with the methods you have listed,
to add in 'demos', wouldn't be a bad idea. But I also understand, we are
all very busy.

Cheers,
    Dan

**Comment by daniele-niero** [comment]
Ok, I've made some tests.

I could achieve the same thing with derive-detach:

```python
def configure(conf):
    # conf.load('compiler_cxx msvs')
    conf.env.CFLAGS = []
    
    common = conf.env.derive()
    common.detach()

    relenv = common.derive()
    relenv.detach()
    conf.setenv('release', relenv)
    relenv["VARIANT_NAME"] = "Release"
    relenv.CFLAGS += ['-O2']
    
    debenv = common.derive()
    debenv.detach()
    conf.setenv('debug', debenv)
    debenv["VARIANT_NAME"] = "Debug"
    debenv.CFLAGS += ['-g']
```

Arguably not as clear as the deepcopy but it works, indeed.
On the other hand, I couldn't manage to  get anything working with the append methods.
I don't see how those can bypass the fact that objects in python are always passed as reference...
Further more they work only with lists... 

Cheers,
   Dan

ita1024:
relenv.append_value('CFLAGS', '-O2')

just works

daniele-niero:
That append a value to a python list and the python list is shared between
Configurations (because list is an object and in python object are always
passed by reference).

So I think we are talking about two different things here.
append_value on a ConfigSet add the same value to the (shared) list in all
the other ConfigSet that are shallow copy of it.

Unless I'm missing something. I will take a look at the source code, I will
answer to myself ;)

ita1024:
Values are shared until they are modified or until detach() is called. Is that what bothers you?

Will you update the pull request or are you expecting me to merge the suitable changes by hand?

daniele-niero:
I think that, because your objection to iteritmes (which makes completely
sense) and the fact that you recently change .derive to return self, so we
can do derive().detach(), I would say that you can ignore my pull request.

only __iter__ would remain, but I think that if you want that you can
always add it later.

Cheers

ita1024:
Thanks!

PR #1595, opened 2015-07-01
Disable regular sys.path when loading tool with specified tooldir.

nikoder:
Currently, when a user attempts to load a waf tool, waf will load any module with the given name, which can be found on ```sys.path```. Specifying a ```tooldir``` does prevent loading from ```waflib.Tools``` and ```waflib.extras```, but not from any location already on the path.

The ability to load waf tools from site packages and other locations on the path may be desired in general, but searching the path, even when ```tooldir``` is specified, breaks my use case of first attempting to load a tool from a specific directory and then falling back to the standard waf tools, if that fails. In my case, a *python module* gets loaded from ```sys.path```, instead of the *waf tool*, which I want.

To me, when specifying a ```tooldir``` the search behaviour should **either** be _identical_ to the normal behaviour, except for _also_ searching the specified directory **or** it should _only_ search the specified directory. This PR implements the latter behaviour.

Alternatively, I could
* implement a version which implements the former behaviour,
* or one which disables searching the standard ```sys.path```, via an optional argument to ```ctx.load()```,
* or one which disables searching the standard ```sys.path``` altogether.

Let me know if you would prefer any of those. They would all solve my problem.

### DIFF
--- a/waflib/Configure.py
+++ b/waflib/Configure.py
@@ -234,7 +234,7 @@ class ConfigurationContext(Context.Context):
 			tmpenv = self.all_envs[key]
 			tmpenv.store(os.path.join(self.cachedir.abspath(), key + Build.CACHE_SUFFIX))
 
-	def load(self, input, tooldir=None, funs=None):
+	def load(self, input, tooldir=None, funs=None, loadFromSysPath=True):
 		"""
 		Load Waf tools, which will be imported whenever a build is started.
 
@@ -260,7 +260,7 @@ class ConfigurationContext(Context.Context):
 
 			module = None
 			try:
-				module = Context.load_tool(tool, tooldir, ctx=self)
+				module = Context.load_tool(tool, tooldir, ctx=self, loadFromSysPath=loadFromSysPath)
 			except ImportError as e:
 				self.fatal('Could not load the Waf tool %r from %r\n%s' % (tool, sys.path, e))
 			except Exception as e:
--- a/waflib/Context.py
+++ b/waflib/Context.py
@@ -621,8 +621,8 @@ def load_module(path, encoding=None):
 	module_dir = os.path.dirname(path)
 	sys.path.insert(0, module_dir)
 
-	exec(compile(code, path, 'exec'), module.__dict__)
-	sys.path.remove(module_dir)
+	try    : exec(compile(code, path, 'exec'), module.__dict__)
+	finally: sys.path.remove(module_dir)
 
 	cache_modules[path] = module
 
--- a/waflib/Context.py
+++ b/waflib/Context.py
@@ -210,9 +210,10 @@ class Context(ctx):
 		"""
 		tools = Utils.to_list(tool_list)
 		path = Utils.to_list(kw.get('tooldir', ''))
+		loadFromSysPath = kw.get('loadFromSysPath', True)
 
 		for t in tools:
-			module = load_tool(t, path)
+			module = load_tool(t, path, loadFromSysPath=loadFromSysPath)
 			fun = getattr(module, kw.get('name', self.fun), None)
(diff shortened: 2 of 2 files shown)


ita1024:
The current behaviour should remain until waf 1.9. Additional arguments are possible.

nikoder:
Ok, understood, makes sense.

Would you like me to target this (or something else) at a different branch, for 1.9?

I'll also be happy to create a PR which implements the behaviour you would like for 1.9 and enables it via an optional arg for 1.8 (if you think 1.9 might be a fair while off, or that it would be valuable to have in general).

ita1024:
Via an optional argument. I would like to avoid changing the behaviour for waf 1.9 too if possible.

ita1024:
Will you provide the option or do you want us to do it?

nikoder:
Thanks Thomas, sorry, I would be happy to implement it, but I don't know  
whether I will be able to return to this for another 3 weeks, or so. If  
you feel this would be worth having sooner rather than later, then please  
feel free to implement it yourself. Otherwise, I shall get to it soon-ish.

nikoder:
Ok, a little earlier than planned, I have pushed my shot at adding an option to solve this (let me know if you'd prefer this in a separate PR, btw - commits are on this one for now).

#### About This Solution

I chose the approach of allowing the user to disable loading from the standard ```sys.path``` altogether (regardless of ```tooldir```), since this provides the greatest control to the user. The default option should preserve the old behaviour exactly, although, of course, I think that the opposite makes more sense as the default, in the long run.

Technical concerns about this solution: I tried to design the modification of the sys.path in a way that is as compatible as possible with potential further modification, by imported code. Anything that tries to do something particularly clever with the regular path may of course fail, though. A solution to this would be to just manually locate and "import" the right file, in the case where the regular sys.path should not be searched. I already have code for the import step in my user code, if you are interested in this.

#### Further Considerations

Specifically, given that waf is intended to be distributed with projects, rather than package managers, I think it makes more sense to just search the known ```waf_dir``` rather than all sys.path (i.e. change the default on the new option, in the long run).

I have already run into the problem that the ```cython``` module from my actual _Cython_ installation gets loaded when I try to load the ```cython``` waf tool, resulting in less than obvious wrong behaviour. This happened because I had forgotten to include the cython tool from extras when I built ```waf```. It would be nice to feel more certain that a lack of error from a ```ctx.load``` means that the tool was actually loaded successfully.

Besides this, I also don't see why providing a ```tooldir``` changes the logic by which the tool is sought, besides either just including the given dir(s) in the search, or just exclusively searching those dir(s).

In the long run, I would thus also vote for a consolidation of those two branches (the ```if tooldir:...``` branch in ```Context.Context.load_tool()```).

ita1024:
"Removed the unused "ctx" argument in waflib.Context.load_tool()" **cannot** be applied due to the API change.
"Add an option to skip searching the regular sys.path when loading waf" is mostly OK, though I would like a better name for the new parameter.
"Ensure restoration of sys.path in Context.load_module()" is probably OK

nikoder:
Ok, I have removed the commit ```Removed the unused "ctx" argument in waflib.Context.load_tool()```.

I don't mind removing ```Ensure restoration of sys.path in Context.load_module()``` as well, and moving it to a new PR or targeting it to a "future" dev branch, if you prefer.

I'm happy to change the parameter name too. How about ```ignoreSysPath``` (with correspondingly negated logic)? I'm afraid I don't have any better naming ideas than that, though.

ita1024:
It is alright, we can find a better name later (with_sys_path?)

nikoder:
No worries, thanks Thomas.

```with_sys_path``` sounds fine to me.

PR #1604, opened 2015-07-27
Handle .. within paths in gccdeps

mattfischer:
The current code handles .. at the beginning of a path, but not within the path


ita1024:
It looks good, but I wonder if
```
while lst and '..' in x:
```
is actually necessary, does your project build with only:
```
while '..' in x:
```
?

mattfischer:
I was unclear what that check was supposed to accomplish, but left it in because it was there before.  My project does seem to build fine without it, so I'll post an updated pull that removes it.

ita1024:
Thank you! :-)

PR #1605, opened 2015-08-03
Added convenience instructions for Windows

brousch:


ita1024:
A waf.bat file already exists: ce8234c396bb246a20ea9f51594ee051d5b378e7

PR #1607, opened 2015-08-12
Fix in-wscript doxyfile customization for Python 3

madprog:
dict.iteritems was removed in Python3, so this code was failing.


ita1024:
"dict.items" works in both python 2 and python 3 so the following is simple enough e9e8df28a401ae80b950c0c693f47767239b8840

madprog:
You are right. However, you create a copy of the list in Python 2, instead of using an iterator.
However the performance increase may not be worth the syntax. Sorry for that.

Issue #1603, opened 2015-07-24
Fortran build without features="fc" fails with "no attribute 'includes_nodes'"


mbeachy:
Without an explicit features="fc", an attempt to build an object file fails with "'task_gen' object has no attribute 'includes_nodes'". This is despite the fact that Tools/fc.py is found and being used.

One fix for this (that works) is to add checks for Fortran file suffixes in [sniff_features](https://github.com/waf-project/waf/blob/92f584fd8f4319d45b9f51e5c76af4197870eda3/waflib/Tools/c_aliases.py#L28) and I can post a pull request to add this. I don't know if this is the _right_ fix, or if some modification to the fc class(es) or simply a more helpful error message is more appropriate.

ita1024:
Adding features="fc" is necessary. There are many cases where the aliases (bld.program/bld.shlib) are unable to detect what the intention is.

Modifying sniff_features can be a possibility for fortran applications, but which linker should be used in programs/libraries mixing fortran, c and c++?

mbeachy:
If a program library mixes all three then the c++ linker is necessary. If it's just fortran/c then the c linker suffices. Linking in some fortran libraries explicitly in either case is often required.

mbeachy:
Eh, maybe I spoke too soon. It looks like we're using fortran for linking shared libs on Linux, c++ on Windows.

ita1024:
Ok. There does not seem to be any change required then.

ita1024:
Re-open if this is not the case.

mbeachy:
An alternate approach to this; improve error reporting: https://github.com/waf-project/waf/pull/1610

I see now that it was probably not intended for bld.program, shlib etc. to be used with Fortran. In that case the error reporting can be modified, but some extension of it would be much appreciated.

PR #1610, opened 2015-08-28
improve error reporting for Fortran programs and libraries

mbeachy:
This change makes bld.program, bld.stlib and bld.shlib throw exceptions
if there is no link task or if a feature of the proper type isn't
specified.

### DIFF
--- a/waflib/Tools/c_aliases.py
+++ b/waflib/Tools/c_aliases.py
@@ -5,8 +5,10 @@
 "base for all c/c++ programs and libraries"
 
 import os, sys, re
-from waflib import Utils, Build
+from waflib import Utils, Build, Errors
 from waflib.Configure import conf
+
+LINKING_FEATURES = ('c','cxx','d','fc','asm')
 
 def get_extensions(lst):
 	"""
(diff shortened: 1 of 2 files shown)


ita1024:
I am not really keen on raising "Exception" :-) How about the more simple 5006e3be5183f07dcb1a31f956002e550fe991ff ?

mbeachy:
The reason I went with raising Exception (of course a more specific exception would have been better, I was a bit lazy there) is that it gives you the line in your wscript where the problem is. (I tried Errors.WafError, which results in a nicely formatted error message, but doesn't help you find the problem. Is there a good way to do this in the Errors framework?)

The solution in https://github.com/waf-project/waf/commit/3e6b3bc5cab2aa038de2d63130c69b97f5cbbc1c _seems_ reasonable, but I thought I agreed with your earlier comments in https://github.com/waf-project/waf/issues/1603 about not knowing which linker was the right one.

Still, 3e6b3bc will prevent the problem of silently exiting when there is no linking feature, and if the wrong linker is being used it won't be too hard to figure out and switch to something else. It seems to meet the main goals I was shooting for.

Issue #1611, opened 2015-08-29
msvc configure checks environment for all installed compilers and targets


shoover:
msvc configure autodetection runs vcvarsall.bat for the product of all architectures and installed compilers. On a system with several versions of Visual Studio this can easily result in 50 checks and take 6+ seconds, even on an SSD. Further, specifying a compiler version or target using the environment keys MSVC_VERSIONS and MSVC_TARGETS does not short-circuit this search.

Since the first compiler and architecture that works is taken, a suggestion is to locate all the combinations but make the actual checks lazy. Most of the time the first one tried will work and the job is done in half a second.

I have a patch that implements the lazy/memoized solution for most compilers but currently breaks winphone and Smartphone. If there is interest in the approach, I can clean it up and email or pull request.

ita1024:
A patch that does not break anything woul dbe best.

shoover:
Pull request #1612

ita1024:
This will take a bit of time to review as several changes unrelated to caching seem to be gathered in one single patch:
* Why change to quiet=Context.STDOUT?
* Why adding MSVC_VERSIONS/MSVC_TARGETS from os.environ to conf.env?
* Changing command-line options can break user scripts. We are not allowed to do that before waf 1.9
* The existing API must be conserved, else the whole matter will have to wait for waf 1.9.
* An improvement can be to create "target" objects that can be evaluated later. Adding new code/API is OK.

shoover:
Thanks for the quick once-over. I will back up and break this down properly
for consideration.

On Aug 29, 2015 12:46 PM, "ita1024" <notifications@github.com> wrote:
>
> This will take a bit of time to review as several changes unrelated to
caching seem to be gathered in one single patch:
>
> Why change to quiet=Context.STDOUT?
The command only checks that the exe runs. The /help output was not used
but fills the build log. I can revert this if the log is otherwise needed.

> Why adding MSVC_VERSIONS/MSVC_TARGETS from os.environ to conf.env?

Just trying to make it behave like other CC variables. Bad idea?

> Changing command-line options can break user scripts. We are not allowed
to do that before waf 1.9
> The existing API must be conserved, else the whole matter will have to
wait for waf 1.9.

No problem. I will fix it.

> An improvement can be to create "target" objects that can be evaluated
later. Adding new code/API is OK.

I had that but it broke the build command because the targets get saved to
the config set. Is there a way to do it and make the objects serializable?

Thanks.
Shawn

[...]

PR #1612, opened 2015-08-29
Lazily check msvc compilers. #1611

shoover:
Fixed a shadowed parameter controlling the return of the architecture or not.

Fixed to collapse supported WinCE platforms by compiler version.

Add OS flags for autodetect options.

Changed --msvc_version to --ms_versions to match expected env key MSVC_VERSIONS
and avoid conflict with the singular key MSVC_VERSION used elsewhere only for
the version number.


shoover:
Tested with multiple msvc targets and WinCE. I can't test Intel. Will test Windows Phone 8.0 once I get it installed.

shoover:
This breaks APIs and command line options and does too much. Drop it.

PR #1614, opened 2015-08-31
Optionally lazily check installed msvc compilers' environments

shoover:
and stop when one is found that matches the user's options. This happens during autodetect only
when the new option msvc_lazy_autodetect or env.MSVC_LAZY_AUTODETECT is set.

These version uses a lazytup class modified from Thomas in an effort to not break all the gather* APIs. Because these get saved in the config set, repr is overridden to display the real tuple value for targets that have been evaluated and a default value for those that have not.

### DIFF
--- a/waflib/Tools/msvc.py
+++ b/waflib/Tools/msvc.py
@@ -30,6 +30,10 @@ or::
 
 Platforms and targets will be tested in the order they appear;
 the first good configuration will be used.
+
+To skip testing all the configurations that are not used, use the ``--msvc_lazy_autodetect`` option
+or set ``conf.env['MSVC_LAZY_AUTODETECT']=True``.
+
 Supported platforms: ia64, x64, x86, x86_amd64, x86_ia64, x86_arm, amd64_x86, amd64_arm
 
 Compilers supported:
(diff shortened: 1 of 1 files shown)


ita1024:
How about reading `conf.options.msvc_lazy_autodetect or conf.env.MSVC_LAZY_AUTODETECT` instead of passing the `lazy` parameter in every method? I am not sure anyone will want to pass lazy to some methods but not to others.

If you do not provide an update I can merge the pull request and refactor this afterwards.

shoover:
No problem, I will update it.

shoover:
This has been tested with winphone, wsdk, wince, and msvc. Is there anyone available who can test intel?

che2:
+1 Thanks for submitting this patch. 
With 3 versions of Visual Studio installed, my project configuration time on Windows went down to about 3 seconds from about 30.
 
I suggest making `--msvc-lazy-autodetect` the default option if the `--msvc-version` option is also set. It doesn't make sense for the user to wait until all versions of Visual Studio installed on the system has been tested if he already knows the version that he wants to use.

ita1024:
We will make this the default option in Waf 1.9

che2:
Thanks!

srouquette:
Thanks for this patch!  :D

PR #1618, opened 2015-09-01
cpplint: check if it has been correctly imported before using it

srouquette:
discovered some problems while working on travis integration.
cpplint: check if it has been correctly imported before using it
boost: add code fragment for Boost.Log, it can be a problem to link it on specific conf (travis)


ita1024:
2 commits in 1? Please make 1 commit for each in the future.

Now, how can that raise a NameError?

ita1024:
Like this: 143ccb7a742d8e425cd617ca5e3456f03846ce47

ita1024:
And why add a dependency on boost_log?

srouquette:
cpplint: ok, i'll rework this patch and make it in another one.

boost: The lib is tested only if the user has specified it in the check_boost (like other lib tested above). If you don't use Boost.Log, it won't be tested.

I had a problem deploying my project on travis, and this could be tested before linking the project.

PR #1619, opened 2015-09-01
boost: check Boost.Log in check_boost if the user uses it.

srouquette:


ita1024:
I hope this is not the start of many more libraries needed in that place :-)

srouquette:
this lib is pretty annoying, I simply removed it after seeing this. On Windows, I link all my libs dynamically. On Arch, I link all dynamically except log and log_setup, or else there are tons of errors. For Travis, it may be related to Ubuntu (old 12.04), I had to update G++ to 4.9 but could only get boost 1.55, or it may be related to the flag I used to build my program. Anyway, it's a pain in the ass and people should know if they can link it or not  :)

PR #1620, opened 2015-09-02
Fixed a local variable shadowing the arch return switch for setup_msvc.

shoover:


shoover:
Tested by tweaking demos/c/wscript like so:

```
--- a/demos/c/wscript
+++ b/demos/c/wscript
@@ -18,7 +18,9 @@
        opt.load('compiler_c gnu_dirs')

 def configure(conf):
-       conf.load('compiler_c gnu_dirs')
+        conf.autodetect(False)
+        conf.load('msvc', funs='no_autodetect')
+       #conf.load('compiler_c gnu_dirs')
        conf.check_cc(fragment="int main() { return 0; }\n")

        try:
```

PR #1622, opened 2015-09-08
c_config: Avoid warnings/errors when checking for C functions

MoSal:
The check for C functions fails with '-Werror' in GCC (5.2).

The cast here triggers an error:
```
  return (int)p;
```

    error: cast from pointer to integer of different size [-Werror=pointer-to-int-cast]

The conversion here triggers another error with '-pedantic -Werror':
```
  p=(void*)(%s);
````

    error: ISO C forbids conversion of function pointer to object pointer type [-Werror=pedantic]

This patch fixes both errors.

Signed-off-by: Mohammad Alsaleh <CE.Mohammad.AlSaleh@gmail.com>

### DIFF
--- a/waflib/Tools/c_config.py
+++ b/waflib/Tools/c_config.py
@@ -25,10 +25,10 @@ cfg_ver = {
 
 SNIP_FUNCTION = '''
 int main(int argc, char **argv) {
-	void *p;
+	void (*p)();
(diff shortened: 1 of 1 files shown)


MoSal:
Was this commit overridden by a forced update?

PR #1625, opened 2015-09-10
Update the pic build flag used for Sun CC

che2:
`-Kpic` had been used to generate position-indenpendent code for Sun CC.
This permits only 2^10 unqiue symbols per shared object on 64-bit
systems.  This is an unreasonably small default. Switching to `-KPIC`
increases the number to 2^29.

According to
http://docs.oracle.com/cd/E37069_01/html/E37074/bjapp.html#OSSCGbjatt,
both `-Kpic` and `-KPIC` have been deprecated, and the `-xcode` option
should be used instead. `-xcode` has been available since at least Sun
Studio 8.

### DIFF
--- a/waflib/Tools/suncc.py
+++ b/waflib/Tools/suncc.py
@@ -53,7 +53,7 @@ def scc_common_flags(conf):
 	v['cprogram_PATTERN']    = '%s'
 
 	# shared library
-	v['CFLAGS_cshlib']       = ['-Kpic', '-DPIC']
+	v['CFLAGS_cshlib']       = ['-xcode=pic32', '-DPIC']
 	v['LINKFLAGS_cshlib']    = ['-G']
 	v['cshlib_PATTERN']      = 'lib%s.so'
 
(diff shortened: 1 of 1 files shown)


ita1024:
Is there a way to know whether the compiler supports this from the version, or do we need a configuration test to avoid breaking existing projects?

che2:
I am not aware of a way to determine when a build option is introduced for Sun Studio CC other than to check the every version of the compiler.

The oldest version of Sun CC that I have access to is version 8 (which was released in 2004), and I have verified that even this old version (and every major version after it)  support the `-xcode` option. 
https://blogs.oracle.com/tatkar/entry/studio_release_names_from_the

It's unlikely for anyone to be using Sun Studio CC versions released prior to 2004, because the standard Oracle support is for 5 years after the release date. http://www.oracle.com/us/support/lifetime-support/index.html

ita1024:
The c++ compiler detection needs similar changes.

How about the following? 68d8d692e7a298ea68c689bd7f59cd6e764b983f

che2:
Thanks. I added a comment to your commit.

ita1024:
Now this is clearer. The change for the c++ compiler has been added in 49f51d547bf8539ef98c1936aa9cf47197f59c8b

PR #1630, opened 2015-10-01
playground/c: update bld methods. Use write_config_header for variants…

shoover:
…so cleanall doesn't delete them.


shoover:
playground/parallel_cmd seems to get the parallel build done with less setup than playground/c, but it was just a couple tweaks to keep the latter working.

PR #1636, opened 2015-10-09
Workaround for missing libpath in python-config

q-p:
The root problem seems to be ```python-config --ldflags``` missing the
library path when Python is built with shared libraries, see
http://bugs.python.org/issue7352.

This used to work in waf-1.7 due to a different detection mechanism.


ita1024:
On which Linux distribution did you see the problem?

ita1024:
I would suggest 1886e770637f80bde293b20ce1e3556be49f860b to avoid adding the flags unconditionally

q-p:
Ok, thanks. This works fine, with the tiny caveat of the multiple log messages on initial failure, i.e.

```
Testing pyext configuration : Could not build python extensions
Testing pyext configuration : yes
```

Issue #1632, opened 2015-10-08
Implicit uselib from packages for Vala tasks


arteymix:
It is very redundant to write the `packages` and `uselib` when the latter can be inferred easily from the first.

Determining the flags should be performed after the package resolution in case of task dependencies (with `use`), so that child task does not have to specify `uselib` for its parent task.

It wouldn't break anything since you have to specify `uselib` in order to build correctly.

arteymix:
The `package` in `check_cfg` refers to the `pkg-config` file whereas the `packages` in `program` refers to the VAPI name. They are typically named the same, but does not reefer to the same concept.

`valac` automatically resolves that by looking at the basename, so it's a sane default for `waf`.

It might be necessary to store additional information in the configure step so that flags for a package can be resolved at the build step and potentially omit the `uselib_store` option.

ita1024:
It is late and I don't really get it. Patches welcome! :-)

arteymix:
@ita1024 there's a little distinction between packages from `check_pkg`  (pkg-config) and Vala packages (VAPI) that have to be taken care of in the implementation

I'll start something now that I have a much better understanding of waf's internals.

ita1024:
I still fail to understand how to enable such a thing and no one is going to do it so i will close the issue now.

arteymix:
It's a my fault: I did not know about `use` back in the time so I ended up with explicit dependencies for all my targets.

PR #1640, opened 2015-10-15
Implicit 'uselib' for Vala packages

arteymix:
This PR is a work-in-progress for issue #1632 

The first commit propagates `uselib` variables in task dependency so that we don't need to specify `uselib` if any dependencies already include the necessary variables.

```python
bld.shlib(target='lib', src='library.vala', uselib='GLIB')

bld.program(target='exe', source='program.vala', use='lib') # uselib is implicit
```
The next step is to infer `uselib` from `packages`.

### DIFF
--- a/waflib/Tools/vala.py
+++ b/waflib/Tools/vala.py
@@ -45,10 +45,14 @@ def init_vala_task(self):
 	"""
 	self.profile = getattr(self, 'profile', 'gobject')
 
+	self.packages = packages = Utils.to_list(getattr(self, 'packages', []))
+	self.use = Utils.to_list(getattr(self, 'use', []))
+	if packages and not self.use:
+		self.use = packages[:] # copy
+
 	if self.profile == 'gobject':
-		self.uselib = Utils.to_list(getattr(self, 'uselib', []))
-		if not 'GOBJECT' in self.uselib:
-			self.uselib.append('GOBJECT')
+		if not 'GOBJECT' in self.use:
+			self.use.append('GOBJECT')
 
 	def addflags(flags):
 		self.env.append_value('VALAFLAGS', flags)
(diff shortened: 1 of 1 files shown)


arteymix:
The vala tool is not using `create_compiled_task`, that's why `uselib` is not being propagated. I'll try to rewrite the task initialization instead.

ita1024:
Err, no that's not the reason.

How about the following branch?
https://github.com/waf-project/waf/commit/7017a194ed82c1514c528e4c0b9599041f3bb337

arteymix:
This should apply to Vala as well, no? https://github.com/waf-project/waf/blob/master/waflib/Tools/ccroot.py#L274 (I added `valac` and it worked)

Isin't `use` for targets and `uselib` for flags?

ita1024:
It already does when calling `bld.program(source='x.vala', ...)` as that translates into `bld(features='c cprogram', ...)`. Adding `valac` in the list of features would require to write `bld(features='valac', ...)` but you did not (a warning will be given when running `waf -v`). Can you elaborate on what you meant by adding "valac"?

The attribute `use` applies for both for targets and flags and enables project to switch between system flags and local targets without changing the build scripts.
If the definition of `uselib` were the only one processed in the patch that I suggested, then it would become impossible to specify the flag order anymore, for example in:
`bld(..., use='foo', packages='gtk+-3.0')` then the package `gtk+-3.0` would be added to `uselib` regardless of what the user wants with `foo`. There is an impact on existing build scripts in both cases (use/uselib) if I am not mistaken.

arteymix:
I didn't know that `use` also works with flags.

The branch looks correct: it's a good thing to append the flags in a way they can be propagated properly with `use`. I will force-push the branch upon this PR.

By 'adding valac' I meant adding it to the `@feature` decorator so that it would propagate the flags, but it's not necessary with the change. The thing is, it would be automatic if `create_compiled_task` was used to create the task.

The only missing thing would be the flag inference from the packages based on what's retrieved from the `configure` step.

ita1024:
I am afraid to apply this for waf 1.8 as it will break the build for some configurations, but the change is in the branch waf-1.9

ita1024:
Use waf 1.9.0

PR #1651, opened 2015-11-07
Xcode6 generator updates and fixes

mimon:
# XCode 6 generator
## Main changes
* The bld(export_headers=..) argument now supports folders (not only files) - this tells XCode to ship these headers with a .framework target

* Added a bld(install='/path/...') argument that tells XCode to copy the target to desired path after build.

* Fixed an issue where a dependency framework target in bld(use='mytarget') was added both in LD_FLAGS and LinkWithFramework Buildphase. Now it's only added to the latter.

Lastly, one commit in this branch made some changes in Configure.find_files but was restored in the newest commit, since the change should not be in this pull request.


ita1024:
Can you rebase your changes please? The commits that change non-xcode files such as "Merge branch 'find_file-fix' into xcode6-gen" or  "Restored Configure.py from an unintended change" will not be taken.

Also:
* `filter` in python3 returns non-serializable data - prefer `list(filter(...))`
* The function `m_hook` does not seem to do anything with .m files

mimon:
> On 07 Nov 2015, at 21:31, ita1024 <notifications@github.com> wrote:
> 
> Can you rebase your changes please? The commits that change non-xcode files such as "Merge branch 'find_file-fix' into xcode6-gen" or "Restored Configure.py from an unintended change" will not be taken.
> 
> Also:
> 
> filter in python3 returns non-serializable data - prefer list(filter(...))
> The function m_hook does not seem to do anything with .m files
> —
> Reply to this email directly or view it on GitHub <https://github.com/waf-project/waf/pull/1651#issuecomment-154746040>.
> 
Thank you for the point about filter(). I’ll fix that.

> The function m_hook does not seem to do anything with .m files
This is a problem I don’t know to solve the proper way. the m_hook is there because the xcode6 tool uses features from cxx and cxx program. If I add .m and .mm files to bld(source=) then it’ll fail with a message ‘No mapping for .m, .mm …’. Being an XCode generator, it should accept obj-c .m and .mm files by default?

ita1024:
I am still not going to take the changes to Configure.py as they are unrelated to xcode.

Your tool is not going to play nice with regular processing of `.h` and `.m` extensions: after just loading xcode6, the files having those extensions will not be processed anymore! Have perhaps the files listed in another attribute than `source` (reserved for extension-based processing) or have your generator ignore errors for the file extensions that are unsupported?

mimon:
Okey, I see. I have refactored the code and replaced the source arg with a new source_files argument that also supports obj-c files so that it won't interfere with other extensions tools. I have also cleaned the commits so that no change is made to Configure.py.

ita1024:
I still see 28 commits for that. Is there any way to reduce to one or two?

mimon:
I got some problems trying to fix and clean the commits up. It's a total mess now, so I have made a new pull request which contains only one commit and the relevant file changes. See #1653.

Closing this.

ita1024:
Yes, that's much better!

PR #1652, opened 2015-11-07
Enable usage of home folder/tilde ~ in Configure.find_files

mimon:
Enables usage of home folder/tilde ~ in Configure.find_files


ita1024:
Why not calling expanduser after joining the paths?

mimon:
Done.

ita1024:
It was a question, not an incentive to do it!

PR #1654, opened 2015-11-16
Add configuration variables to Dist context.

jpedrick:
I suspect this functionality should be part of Context.Context, but for now I hacked it into Dist.

### DIFF
--- a/waflib/Scripting.py
+++ b/waflib/Scripting.py
@@ -344,11 +344,71 @@ class Dist(Context.Context):
 	fun = 'dist'
 	algo = 'tar.bz2'
 	ext_algo = {}
+	variant = ''
+
+	def __init__(self, **kw):
+		super(Dist, self).__init__(**kw)
+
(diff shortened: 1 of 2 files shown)


ita1024:
Well, no:
1. This makes the dist command depend on the configuration ()
2. The code is merely copy-pasted
3. The changes affect incompatible modules (qt5?)

jpedrick:
Totally fine with the reject. 
#3) Was simply a bugfix in qt5.py. I should have isolated the change.
#2) Yes, meant as a proof of concept to further discussion.
#1) What's wrong with the dist command depending on your configuration?

ita1024:
1. This is a new dependency that cannot be introduced in waf 1.8. That will have to wait 1.9 at least.
2. Ok.
3. The options should be loaded, but if you want to load your own options, the same change should be applied to qt4. e66cf563ce37385f0302d9bdda6b881735dc85ac

jpedrick:
Sure, makes sense to wait until 1.9.

PR #1658, opened 2015-11-20
Add Node.read_json()

mattyclarkson:
Reading a JSON file and converting the contents to native strings isn't
that trivial. This patch adds Node.read_json() that parses a file as
JSON and then converts the returned objects into the native strings types
for Python 2 or 3. This is due to the JSON module always returning unicode
strings.

We use this code a lot in our tool chain. Thought others might find it useful.


ita1024:
While this may be useful, the implementation raises a few questions:
1. How about write_json?
2. What is that type() line doing?!
3. Is the unicode conversion for python2?
4. I want an example in the form of a playground folder to show how this works

mattyclarkson:
1. ooo, that's a good idea! I could implement that
2. it recursively calls convert on all the data inside the iterable
3. the unicode conversion is for python2
4. I can provide that.

Thinking about this with fresh eyes, I can use the JSONDecoder class to do the conversion of strings to native strings (byte string on python2, unicode strings on python 3) without the `convert` function. Let me revisit this and I'll implement the things you said about.

Thanks for your comments!

PR #1662, opened 2015-11-30
add explicit licensing information

darekp:
Currently I don't see easy to find and explicit information about the licensing the waf code. By adding COPYRIGHT new users will be able to find the license info much quicker. What do you think about this change?


ita1024:
The licensing information is right in the waf file:
https://github.com/waf-project/waf/blob/master/waf-light#L5

PR #1666, opened 2015-12-08
some fixes; editable for vim with different defaults; Win: WiX and satellite_assembly in extras

rpuntaie:
... plus an example using them in playground: stpl_c_py_cs_satellite_wix
Appart from WiX it should also be possible to make this work on linux with mono, but that is a todo (rpath, ...). Please, check the commit messages for more.


ita1024:
My vim does not have problems with tabs or spaces. Changing all the files in the project just for that seems a bit excessive...

Hidden in the middle of all that there are changes to `find_binary` that are unexplained, new commented lines in `waflib/Context.py`, a few API changes, variable renames for no reason...

At least:
  * case insentive extension matching almost certainly breaks C extension for c++.
  * calling mkdir in `node.write` looks convenient but changes API, complicates exception handling and is inconsistent with `Utils.writef`.

Could you make 1 pull request for what is only C# tool enhancements after taking out all the "minor fixes"?

rpuntaie:
Please let's discuss these changes in a little more detail before discarding. I've invested some effort for the good of the project.

> My vim does not have problems with tabs or spaces. Changing all the files in the project just for that seems a bit excessive...

I suppose many python developers out there that use vim will have settings that follow https://www.python.org/dev/peps/pep-0008/#indentation, which suggests 4 spaces. I do. 
The change in this regard, which I implemented in the pull request, is to add the line `# vim: tabstop=4 noexpandtab` at the top.
This will override the developers default settings and comply to the file's convention to use tabs instead.

This is a very convenient change of all potential contributers that use vim.

> Hidden in the middle of all that there are changes to find_binary that are unexplained, new commented lines in waflib/Context.py, a few API changes, variable renames for no reason...

The change next --> nxt was because I often execute line by line in vim, in this case I later realized that the default python next got overridden. But I undid this.
I have commented disputed further changes in code in greater detail (You can remove the comments, when you merge.)

> case insensitive extension matching almost certainly breaks the ".C" extension for c++.

I'm not sure what you refer to. Are you referring to the fnmatch change?

> calling mkdir in node.write looks convenient but changes API, complicates exception handling and is inconsistent with Utils.writef.

Utils.writef is a separate function. There is no need for them to have the same behavior or the same interface.

ita1024:
> I'm not sure what you refer to. Are you referring to the fnmatch change?

Yes.

rpuntaie:
>> case insensitive extension matching almost certainly breaks the ".C" extension for c++.

>I'm not sure what you refer to. Are you referring to the fnmatch change?

    if name.endswith(k) or fnmatch(name,k):

This means that ``fnmatch`` is only called if ``name.endswith`` does not apply,
i.e. actually no change for all extensions in waf.
Note also that the file name description intended for ``fnmatch`` must have a * in it, like "*.xxx" and not ".xxx".
A "*.xxx" will never work for ``endswith()`` and ".xxx" will never work for ``fnmatch()``. There is no overlap.

"*If the operating system is case-insensitive*, then both parameters will be normalized to all lower- or upper-case before the comparison",
i.e. in Linux it is not case insensitive and in Windows it should be.

So this is not a change, it is an extension.

ita1024:
There are too many changes crammed in a few patches and they include API changes (return types), functionality changes (new command arguments), changes that definitely affect performance (fnmatch) and changes that affect too many files at once.

Composite extensions such as `*.foo.txt` also pose a design issue. It is much better to declare the source files in a particular attribute instead of trying to share `.txt` files with all other builders:
```python
@feature('wix')
@before('process_source')
def process_wix_sources(self):
    nodes = self.to_nodes(self.wix_sources)
    for x in nodes:
         pass # only '*.??.txt' and '*.??.resources' nodes are expected here
    # if not re-using self.source, just set "self.source=[]" here
```

The changes to `resx.py` have been applied in bdca245588005a7fa0bed679bbac31a8fe25db99. The rest of applicable changes will wait for waf 1.9. The wix support may get there earlier if it is decoupled from the extension processing changes as described above.

ita1024:
The tool `wix.py` has been added as 95114ff2057a6d652bd0a53309efd3208d915a1d.

rpuntaie:
Regarding wix.py: 

- You did not merge the newest version, which uses ``copy.copy``, because python 2.x does not have a list ``copy``
- There should be a playground project using wix.py. ``playground\stpl_c_py_cs_satellite_wix`` would be one. Please consider merging.

Regarding satellite_assembly and fnmatch:

- I've used the feature method for satellite_assembly and removed fnmatch. Maybe you can merge this now.

Regarding the other changes:

- They are really helpful. Please let's continue discussing them.

rpuntaie:
I've removed the vim related lines again.
The playground playground\stpl_c_py_cs_satellite_wix project builds on Windows.
Can you help me build it on Linux such as to make it cross-platform, as a test for waf?

ita1024:
This is better. With a few tweaks the non-wix part seems to run on Linux at least:

```
D_LIBRARY_PATH=$PWD/../build/api/: PATH=$PATH:$LD_LIBRARY_PATH waf configure build test --stubs
Setting top to                           : /home/user/waf/playground/stpl_c_py_cs_satellite_wix/src 
Setting out to                           : /home/user/waf/playground/stpl_c_py_cs_satellite_wix/build 
Checking for program 'py.test'           : /usr/bin/py.test 
!USING STUBS!
Checking for 'gcc' (C compiler)          : /usr/bin/gcc 
Checking for 'g++' (C++ compiler)        : /usr/bin/g++ 
Checking for program 'csc, mcs, gmcs'    : /usr/bin/mcs 
Checking for program 'resgen'            : /usr/bin/resgen 
Checking for program 'al'                : /usr/bin/al 
'configure' finished successfully (0.031s)
Waf: Entering directory `/home/user/waf/playground/stpl_c_py_cs_satellite_wix/build'
[ 5/19] Compiling api/funi.py.stpl
Waf: Leaving directory `/home/user/waf/playground/stpl_c_py_cs_satellite_wix/build'
'build' finished successfully (0.033s)
('running test in ', '/home/user/waf/playground/stpl_c_py_cs_satellite_wix/build/api')
/home/user/waf/playground/stpl_c_py_cs_satellite_wix/build/api/test_funiout: 0
1
2
2
/usr/bin/py.test test_funi.pyout: ============================= test session starts ==============================
platform linux -- Python 3.4.1 -- py-1.4.22 -- pytest-2.6.0
collected 4 items

test_funi.py ....

=========================== 4 passed in 0.02 seconds ===========================
'test' finished successfully (0.125s)
```

Now there are a few things that I do not really want to merge here:
  * The extension in `find_binary` is meant for finding programs in a cross-platform way: searching for `gcc` will return `gcc` on Linux and `gcc.exe` on windows
  * The API changes in Node.py will not be applied at this time
  * The additional flags in `cs.py` may cause regressions, this will wait for Waf 1.9

rpuntaie:
Thanks a lot for making playground/stpl_c_py_cs_satellite_wix build on linux and for including it.
I've reverted the find_binary change.

ita1024:
Waf 1.9 will accept expressions of the form:

```python
@extension(re.compile('.*[.]..[.]resources'))
def function_name(self, node):
```

I am really reluctant to change the Blue color: if it is not readable then it is a problem with the terminal settings. It would also break existing scripts unnecessarily.

I think that I have merged all the things that needed to be merged into the branch waf-1.9. If this is not the case make a new pull request.

Issue #1668, opened 2015-12-10
Provide a way to know the variant build node at configuration time


guludo:
Sometimes it's necessary to know where will be the build node for the current env during configuration time. I'm talking about a scenario when we have variant builds and different envs for each (through the function `setenv()`).

One example is my case: I'm building an external tool with an external build system, and I need to tell the external build system configuration where it should build its stuff. :-)

ita1024:
The folder can be obtained easily by looking up the output directory and the name of the variant folder. No change is deemed necessary.

PR #1669, opened 2015-12-10
Add vbldnode to ConfigurationContext

guludo:
Sometimes during configuration is necessary to know where the variant build is
for the current env. One example would be configuring external build system.

This addresses issue #1668

### DIFF
--- a/waflib/Configure.py
+++ b/waflib/Configure.py
@@ -99,6 +99,15 @@ class ConfigurationContext(Context.Context):
 
 	env = property(get_env, set_env)
 
+	def get_vbldnode(self):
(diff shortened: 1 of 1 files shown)


guludo:
I'm not sure if `vbldnode` is a good name. I think it's short and easy to remember.

guludo:
Should I close this, since it hasn't received a response?

ita1024:
We will not apply this pull request for the following reasons:
  * The needs are unclear
  * Project-specific code can be added easily through @conf methods
  * Similar changes would be required in BuildContext, so this is creating an inconsistency

PR #1671, opened 2015-12-11
c_config: Don't append library names uniquely

MoSal:
Sometimes, we need to pass a library name to a linker twice to handle
 circular dependencies.


ita1024:
Applying this can cause additional regressions. Can this wait for waf 1.9?

MoSal:
@ita1024 sure.

ita1024:
Applied as cbaab2f1b0e95753e34e3a49a52c31721312a527 in waf-1.9

Issue #1672, opened 2015-12-16
waf fails to extract ifort compiler version on Windows (repeat of #1473)


apthorpe:
I'm running into the same problem detailed in (closed) bug #1473 where waf find the Intel Fortran compiler (ifort) on Windows 7, executes it to get the version info, but fails to parse the results, thus claiming it can't find the compiler.

A minimal test case is attached including the build/config.log which shows the proper detection of the compiler binary and the output from running it to get version info.

[winifort_failure_report.zip](https://github.com/waf-project/waf/files/64525/winifort_failure_report.zip)

The important part of log file is at the end:

    ['c:\\Program Files (x86)\\Intel\\Composer XE 2013 SP1\\bin\\ia32\\ifort.exe']
    err: Intel(R) Visual Fortran Compiler XE for applications running on IA-32, Version 14.0.2.176 Build 20140130^M
    Copyright (C) 1985-2014 Intel Corporation.  All rights reserved.^M
    ^M
    ifort: command line error: no files specified; for help type "ifort /help"^M
     
    from C:\Users\Apthorpe\Downloads\inst\winifort-0: could not determine the compiler version ['c:\\Program Files (x86)\\Intel\\Composer XE 2013 SP1\\bin\\ia32\\ifort.exe']
    not found
    from C:\Users\Apthorpe\Downloads\inst\winifort-0: could not configure a Fortran compiler!

I haven't been able to test a fix, but if it turns out that the ifort version detection regex is failing, you may want to change waflib/Tools/ifort.py from

    @conf
    def get_ifort_version(conf,fc):
            version_re=re.compile(r"Intel[\sa-zA-Z()0-9,-]*Version\s*(?P<major>\d*)\.(?P<minor>\d*)",re.I).search

to

    @conf
    def get_ifort_version(conf,fc):
            version_re=re.compile(r"Intel[^A-Z0-9].*Version\s*(?P<major>\d*)\.(?P<minor>\d*)",re.I).search

If the re.I option is specified, there's no need to look for both [a-z] and [A-Z]; the regex can be loosened a little so if Intel randomly sprinkles more punctuation in their difficult-to-gracefully-obtain version info next year, hopefully the regex won't start failing again.

ita1024:
> hopefully the regex won't start failing again

Is there any command that we could use to display the version in the same way as [get_cc_version](https://github.com/waf-project/waf/blob/master/waflib/Tools/c_config.py#L1011)?

apthorpe:
[...]

PR #1674, opened 2015-12-23
Ignore error codes from cmd_and_log in getoutput for Fortran compilers

haraldkl:
As discussed in #1672 here is a patch to ignore error codes from Fortran compilers when obtaining the version string.

### DIFF
--- a/waflib/Tools/fc_config.py
+++ b/waflib/Tools/fc_config.py
@@ -346,11 +346,16 @@ def getoutput(conf, cmd, stdin=False):
 	try:
 		out, err = conf.cmd_and_log(cmd, env=env, output=0, input=input)
 	except Errors.WafError as e:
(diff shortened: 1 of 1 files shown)


ita1024:
Yes. There is one case in `cmd_and_log` where `WafError` can be raised without providing `e.stdout`/`e.stderr` though.

haraldkl:
So, now I am checking, whether thes attributes exist. In the other case the executable does not exist and it should be fine to raise the errror.

ita1024:
Thanks!

ita1024:
It looks like there are indentation issues with python3   https://waf.io/cilog/?jobnum=238&jobname=waf-pyflakes

haraldkl:
Ah, sorry. No clue how this happened...

PR #1683, opened 2016-01-14
add empty check for inst.exec_install_as()

zonglu1988:
similar as exec_install_files(), exec_install_as() should check empty for self.inputs.

It causes exception if the inst is not posted.


ita1024:
Why? A lack of inputs there indicates a mistake: what is the point of installing a file without any source file to create it?

drahnr:
Depends on the output generated, if it allows easily relate the exception to the input being `None`/`""`/...

ita1024:
Setting an empty value in the inputs is a mistake. This change cannot be applied.

PR #1684, opened 2016-01-14
Fixing version flag for NAG compiler.

aradi:
The nagfor compiler (as of current version 6.0) uses according to the man page the option `-V` to query compiler version. Waf uses `-v` instead, where nagfor expects also a source file name at the command line and generates an exit code != 0 when not present. This caused in waf 1.8.17 failure when detecting nagfor. Although recent patches in `fc_config.py` work around this by accepting non-zero exit codes, one should probably still use the correct flag.

### DIFF
--- a/waflib/extras/fc_nag.py
+++ b/waflib/extras/fc_nag.py
@@ -36,7 +36,7 @@ def get_nag_version(conf, fc):
 	"""Get the NAG compiler version"""
 
 	version_re = re.compile(r"^NAG Fortran Compiler *Release *(?P<major>\d*)\.(?P<minor>\d*)", re.M).search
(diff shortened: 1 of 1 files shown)


ita1024:
@haraldkl please check

haraldkl:
Yes, it should be -V instead of -v. Works for me after the change. The regex for the version string itself still seems to be valid.

Here's the output in the config log I get:

    ['nagfor', '-V']
    err: NAG Fortran Compiler Release 6.0(Hibiya) Build 1057
    Product NPL6A60NA for x86-64 Linux
    Copyright 1990-2015 The Numerical Algorithms Group Ltd., Oxford, U.K.

Thanks.

PR #1692, opened 2016-01-25
No-op in netcache run/post_run if nocache is true

i-tub:
When a task class has a true value for the nocache attribute,
setup_nocache doesn't decorate the run() and post_run() methods of the
class. In most cases, however, that's not enough to disable caching
because the base Task class is still modified; any class based on Task
which doesn't override these methods will still use caching even if it
sets nocache to True. This is solved by having the decorated versions of
run() and post_run() do nothing except call the original version of the
method when the task object has a true value for self.nocache.

### DIFF
--- a/waflib/extras/netcache_client.py
+++ b/waflib/extras/netcache_client.py
@@ -326,6 +326,8 @@ def make_cached(cls):
 
 	m1 = cls.run
 	def run(self):
+		if getattr(self, 'nocache', False):
+			return m1(self)
 		if self.can_retrieve_cache():
 			return 0
 		return m1(self)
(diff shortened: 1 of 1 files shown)


ita1024:
Setting nocache on a subclass should not affect the upper classes.

i-tub:
My implementation doesn't affect the  upper classes, because the new ifs inside run() and post_run() check self.nocache, not cls.nocache. An object of class Task, or most of its descendants, won't have this attribute, so it won't skip the caching. But an object of class "MyTaskWithNoCaching(Task)" will have self.nocache == True and will skip the caching.

ita1024:
Yes, correct :+1:

PR #1695, opened 2016-02-02
make Task constructor aware of create_task arguments

dffischer:
All unrecognized keyword arguments to the task generator function `create_task` are merged into the `__dict__` of the created task. This is a useful feature to pass additional information per task.

However, by inforcing these arguments with `task.__dict__.update` after the task object has already been created, they completely bypass the task constructor. So task subclasses have no direct possibility to perform any initialization depending on values passed to them this way.

The patches in this pull request move the flow of arguments so that they pass the constructor. Task subclasses can then inspect arguments passed to them via `create_task`, for example in their `__init__` method, and react accordingly.

The change should not break any existing API, as it only extends the argument list of `create_task`, which should not affect any already working code. I also took care not to pass more arguments to any related functions, especially not to clutter the `__dict__`s of all tasks with a heap of variables or overwrite anything important.


ita1024:
Changing `def __init__(self, *k, **kw):` to `def __init__(self, inputs=None, outputs=None, *k, **kw):` seems to break existing code of the form "Task(obj1, obj2, inputs=[])".

dffischer:
I moved this `*k` around and eventually it ended up in the wrong place. When it precedes the new keyword arguments, it should not swallow any positional arguments. I amended the respective commit.

Alternatively, both arguments could be left out entirely, their presence instead checked through `if 'inputs' in kw`, if you consider this more suited for waf.

ita1024:
Such function signatures are not allowed in Python2:

```python
def foo(self, *k, inputs=None, outputs=None, **kw):
        print("hello", self)
foo('56')
```

ita1024:
We have decided that we will not apply such changes as they bring unnecessary complexity and breaks existing functionality.

PR #1698, opened 2016-02-07
Make D dependency scanner aware of package.d files

SirNickolas:
When importing `path.to.module` in D, both `path/to/module.d` and `path/to/module/package.d` should be tried.

### DIFF
--- a/waflib/Tools/d_scan.py
+++ b/waflib/Tools/d_scan.py
@@ -108,16 +108,17 @@ class d_parser(object):
 		:param filename: file to read
 		:type filename: string
 		"""
-		found = 0
+		file_path = filename.replace('.', '/')
 		for n in self.incpaths:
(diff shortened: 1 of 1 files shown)


ita1024:
A minimal example is needed, can you please provide one file or two for `demos/d` that shows how the dependency must be obtained?

ita1024:
I wish you could provide an example :-/

SirNickolas:
Sorry, I was fighting the errors while trying to make compilation with headers (.di) work. It came out that the D compiler had a [bug](https://issues.dlang.org/show_bug.cgi?id=15676) in it.

I'll provide full D support (as much as it can be done for now) in a few days. Sorry again for the inconvenience.

ita1024:
This is unclear. If you do not provide the example I am afraid that I will have to revert the change.

SirNickolas:
When I finish, there are examples for every use case. However, you can revert it for now.

PR #1700, opened 2016-02-11
Output command as string when task fails

guludo:
Several tasks use non-shell commands, which are usually represented as lists.
When a task fails, it's a good idea to output the command as a string instead
of a list, so that it's easy to copy/paste the command and debug it.

### DIFF
--- a/waflib/Task.py
+++ b/waflib/Task.py
@@ -351,6 +351,8 @@ class TaskBase(evil):
 		:rtype: string
 		"""
 		msg = getattr(self, 'last_cmd', '')
+		if not isinstance(msg, str):
+			msg = ' '.join(repr(x) if ' ' in x else x for x in msg)
 		name = getattr(self.generator, 'name', '')
 		if getattr(self, "err_msg", None):
 			return self.err_msg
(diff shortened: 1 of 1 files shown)


guludo:
Ping: just updated the commit, so that `%s` is used instead of `%r`.

ita1024:
The change suggested makes things harder to troubleshoot in the following ways:
 * As the special characters and invalid data will be made invisible (\a, \v, etc)
 * The command printed is not necessarily valid as a shell command (unescaped spaces)
 * The command will have shell-specific quirks ("#" characters)

If you want such a thing, use `waflib/extras/print_commands.py` in your project.

lucasdemarchi:
@ita1024 would it suffice to escape it before printing? The point here that is not fulfilled by `waflib/extras/print_commands.py` is that I want the command that failed, not all the commands being executed.  It's particularly helpful when you are compiling with -j<big-number> and something fails. Re-running with -j1 (and -v) or manually backtracking to which command really failed is a pain. Can't the build system help with it?

lucasdemarchi:
@ita1024 btw, I'm not saying the commit above is correct (I think it still misses a call to an escape function). I'm just trying to justify the change.

ita1024:
The change suggested makes things much worse by displaying incorrect debugging information. Escaping shell-dependent, and knowing which shell is used (cmd.exe/msys/bash/ksh) is not really possible in advance.

If you are debugging your system, import `waflib/extras/print_commands.py` in **your own** project. I am really reluctant to ship such a thing by default as it will cause problems in the long run.

ita1024:
Or not even in the long run. Writing a python script might be relevant but that should support the msvc response files and use the relevant python interpreter.

lucasdemarchi:
@ita1024 so if the patch is fixed and support to msvc added, would it be acceptable? I really don't like to diverge from upstream on functionality that may be useful to other projects.

PR #1701, opened 2016-02-11
fix potential concurrency issue for Waf tool cpplint if using parallel jobs

danbar:
If parallel jobs are used for the Waf tool cpplint, a potential concurrency issue is triggered.
The reason seems to be that the function ProcessFile (file waflib/extras/cpplint.py, line 205) shares the global module state _cpplint_state (file waflib/extras/cpplint.py, line 200) across parallel jobs.


ita1024:
If there are multiple cpplint tasks, then such tasks will all inherit the same data from the last version of self.env.CPPLINT_FILTERS found?

danbar:
As far as I understood, each cpplint task should get the same version of self.env.CPPLINT_FILTERS. 
The problem seems to be that ProcessFile is changing the filters in _cpplint_state. If multiple tasks are running concurrently, they are potentially influencing each other. I was able to resolve this issue in our project by shifting the line _cpplint_state.SetFilters(self.env.CPPLINT_FILTERS) to the constructor.

ita1024:
`self.env.CPPLINT_FILTERS` is not necessarily the same for each cpplint task.

danbar:
Would it have any influence, if self.env.CPPLINT_FILTER is different for each cpplint task?
I think the (potential) problem is that _cpplint_state.SetFilters(self.env.CPPLINT_FILTERS) is located in the run method and not in the constructor. Or am I missing something here?

ita1024:
Yes, it will: the state will be set by the most recently created task.

danbar:
Ok, I understand. But the state is anyhow shared across tasks, because _cpplint_state is a global state of the module. Please see line 848 in https://github.com/google/styleguide/blob/gh-pages/cpplint/cpplint.py

ita1024:
The build order is therefore incorrect and the change suggested should not be applied.

danbar:
I agree. The thing is that I observed that the build order influences cpplint in the sense that the filters get "corrupted" if I am using parallel jobs. The (potential) explanation was that ProcessFile shares a global module state, which is changed during the function call.

ita1024:
Try 33c72e06915d7f40df33ccb639beda9d3b11208a

danbar:
Thx. I will try it out tomorrow and will let you know if it works out.

danbar:
Tried it out and seems to work!

PR #1703, opened 2016-02-15
vala: Initialize 'VALAFLAGS' with OS flags

arteymix:


ita1024:
I am really unsure about `self.env.append_unique('VALAFLAGS', ['-C'])`. If the vala flags are provided in the environment, why change them?

arteymix:
The tool essentially rely on C code generation (triggered by the `-C` flag). It might be nicer to put that into the task though.

The reason OS flags are applied is because it's handy to pass `--debug` or `--vapi-comments` flags to customize the build.

ita1024:
I missed the `append_unique` there. Looks good!

arteymix:
Nice, thanks, this is a game changer for me!

PR #1709, opened 2016-02-26
Boost module enhancements and fixes

cawka:
- Output detected version of boost in dot-form (e.g., 1.56.0, instead of 1_56)
- Fix Boost.Log library detection:
  * when linking to shared library, BOOST_LOG_DYN_LINK needs to be defined
  * when linking to non-multithreaded version, BOOST_LOG_NO_THREADS needs to be defined

  (see http://www.boost.org/doc/libs/1_60_0/libs/log/doc/html/log/installation/config.html)

### DIFF
--- a/playground/boost_log/main.cpp
+++ b/playground/boost_log/main.cpp
@@ -1,0 +1,11 @@
+#include <boost/log/trivial.hpp>
+
+int main(int, char*[])
+{
+    BOOST_LOG_TRIVIAL(trace) << "A trace severity message";
+    BOOST_LOG_TRIVIAL(debug) << "A debug severity message";
+    BOOST_LOG_TRIVIAL(info) << "An informational severity message";
+    BOOST_LOG_TRIVIAL(warning) << "A warning severity message";
+    BOOST_LOG_TRIVIAL(error) << "An error severity message";
+    BOOST_LOG_TRIVIAL(fatal) << "A fatal severity message";
+}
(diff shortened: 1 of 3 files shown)


cawka:
Unfortunately. I have discovered issues with my patch that affects some Linux systems. Will try to patch the patch.

Just a small background.  Boost.Log for some reason decided to have completely different internal naming when the library compiled in different way:

- `v2s_st` internal namespace for static single-threaded version of library
- `v2s_mt_posix` namespace for static multi-threaded with posix, `v2s_mt_nt6` for dynamic on recent windows, `v2s_mt_nt5` for other windows, `v2s_mt` for other multithreaded
- `v2_st` namespace for dynamic single-threaded version
- `v2_mt_posix` namespace for dynamic multi-threaded with posix threads, `v2_mt_nt6` for dynamic on recent windows, `v2_mt_nt5` for other windows, `v2_mt` for other multithreaded

When compiling, it is important to specify proper defines for the linked version of the library.  And here is the problem: there is no easy way to figure it out.  Static/shared is relatively easy (though not on OSX platform).  Multithreaded/single-threaded is very ambiguous.

I tried to add some logic based on .so/.dylib/.a name (presence or absence of `-mt` suffix).  However, this worked only on OSX platforms and completely failed on recent Debian---the library is compiled as multithreaded but no `-mt` prefix added.

The only way I can think of to solve the problem, is to try linking multiple times with different options.

PR #1713, opened 2016-03-02
Add cmd_kw attribute to Task.Task

guludo:
Useful if the user still wants to use the benefits of run_str while tweaking
how exec_command() is going to be called.


lucasdemarchi:
@guludo maybe give the example in which we need this?

ita1024:
* Passing `env` in `cmd_kw` will cause errors.
* No rebuilds will happen when the contents of `cmd_kw` change (a mistake we already made with the `env` parameter)
* This change is not really necessary as subclasses can redefine the method `exec_command`.
* There are calling arguments (input and timeout for example) that the change does not support and will hinder in the future
* There are other plans for waf 1.9, so we are not going to include this.

We cannot apply this, sorry.

PR #1715, opened 2016-03-09
Fix extra lines in output logs

DigitalDan05:
Replacing '\r' with '\n' in log output on Windows causes an extra blank
line between every line of output.

This was originally added in commit a29f775851a71c4e3516c8531f3b7839d8213f88.

I'm not sure if a replacement is even necessary since removing the line entirely also fixes the problem.

### DIFF
--- a/waflib/Logs.py
+++ b/waflib/Logs.py
@@ -200,7 +200,7 @@ class formatter(logging.Formatter):
 			c2 = getattr(rec, 'c2', colors.NORMAL)
 			msg = '%s%s%s' % (c1, msg, c2)
 		else:
-			msg = msg.replace('\r', '\n')
+			msg = msg.replace('\r\n', '\n')
 			msg = re.sub(r'\x1B\[(K|.*?(m|h|l))', '', msg)
 
 		if rec.levelno >= logging.INFO: # ??


ita1024:
Are you calling waf in cmd.exe with "--color=no" and a the progress bar?

DigitalDan05:
I am calling waf in cmd.exe but redirecting the output to a log file.  I am not using --color=no or any progress bar options.  I am willing to test those if I know what combination to look for.

ita1024:
It seems that this works as initially intended. @zougloub, any comment on this?

DigitalDan05:
The behavior changed between Waf 1.7.16 that I was using and 1.8.0

ita1024:
@DigitalDan05 The change suggested has the same effect as commenting the line out. I cannot quite remember the reason why the \r had to be replaced but I would like to know before making a change.

ita1024:
Single `\r` characters must be removed, and this can be achived by means of a regexp: 79dc2698fe7c4d4c2e9a8df0b80c317e4959ec92

DigitalDan05:
I just tested your change and verified that it fixes my issue.  Thank you.

PR #1718, opened 2016-03-21
Use stash/revert/commit pattern for transactional configuration

guludo:
Apply that pattern when relevant. That allows correct behavior if tools are
loaded inside a transaction.

Best regards,
Gustavo Sousa


ita1024:
Applied in waf-1.9 as d46f5410994db98718abe74bb4dd192e39e81fdf. There is no need for this in waf 1.8 and it is more risky to apply it there.

ita1024:
Use waf 1.9.0

PR #1723, opened 2016-04-07
Remove unused libmarkers from the Fortran compiler command.

haraldkl:
Diff to the recent discussion on the mailing list about removing unnecessary library markers from the Fortran command line. This is the variant I am using, which is basically a verbatim copy of your suggestion.


ita1024:
The appropriate location for this is `waflib/Tools/ccroot.py`

haraldkl:
OK, but only for the fc features or would you like this also for the c tools?

ita1024:
This will be added for c/cxx tools in waf 1.9. I will try to have a look this weekend.

ita1024:
With this change `demos/fortran/` will not configure anymore: with `-Wl,-Bstatic` given without `-Wl,-Bdynamic` ld cannot find -lc and -lgcc_s.

haraldkl:
Hm, on my Mac it worked for gfortran.

haraldkl:
Also on my Linux box it seems to work.

ita1024:
The reason is that the flags are not really removed. I believe that the flag order should be set to:
```
@after_method('apply_link', 'process_use', 'propagate_uselib_vars')
```
Conditional flags can be set in waf 1.9 to achieve the same effect, try setting this in fc.py ("?|" was added today):
```
run_str = '${FC} ${LINKFLAGS} ${FCLNK_SRC_F}${SRC} ${FCLNK_TGT_F}${TGT[0].abspath()} ${RPATH_ST:RPATH} ${FCSTLIB_MARKER?STLIB|STLIBPATH} ${FCSTLIBPATH_ST:STLIBPATH} ${FCSTLIB_ST:STLIB} ${FCSHLIB_MARKER?LIB|LIBPATH} ${FCLIBPATH_ST:LIBPATH} ${FCLIB_ST:LIB} ${LDFLAGS}'
```

haraldkl:
Sorry for my late reply. It looked to me, like the libmarkers were removed from the link line in the fortran demo. However, I also had to include fcprogram_test in the list of features to get this done for all tests. 

The Conditional flags seem to work nicely in my settings. It would be fine for me to go ahead and use that run_str for fcprogram class. Thank you very much!

ita1024:
Does demos/fortran still configures and build on your Linux box after the flags are removed? I am getting errors here at least.

haraldkl:
Indeed, now it also fails for me on Linux:

    ['/usr/bin/gfortran', 'main.c.5.o', '-o/user/waf/demos/fortran/build/.conf_check_4b3a6bd5319d94880390d4e2be8d7347/testbuild/app', '-Wl,-Bstatic', '-L.', '-ltest']
    err: /usr/bin/ld: cannot find -lgfortran
    /usr/bin/ld: cannot find -lgcc_s
    /usr/bin/ld: cannot find -lquadmath
    /usr/bin/ld: cannot find -lgcc_s
    /usr/bin/ld: cannot find -lgcc_s

So systems which do provide these libs as static ones where the linker finds them, work fine, but usually it is not working. The funny thing is, that the current solution with the after_method decorator seems to work for me on all platforms. Seems like it does not always remove the markers? Maybe I was just lucky so far...

Too bad, the conditional flags would be a very nice solution.

ita1024:
Yes, you were.

ita1024:
I think I will close this unless there is an obvious workaround.

haraldkl:
Yes please, and thanks for looking into it!

PR #1725, opened 2016-04-10
javac in javaw waftool can "run_after" any task

jbendes:
Originally the javac feature in the javaw waftool was unable to track
"run_after" dependencies that weren't 'jar_task' tasks. Now it can be
configured to run after any task.

Now the javac tool has a unique uid. It didn't before because uid
calculation is based off of inputs and outputs, which javac didn't
register until after it was done waiting on it's run_after tasks to
finish


jbendes:
Completely open to changing the fields included in the uid hash calculation. Right now it's only using srcdir. If you think we should add more to that, let me know and I'm happy to make the change

ita1024:
A few questions:

1. The uid change seems that you are trying to compile the same class files more than once. In this case, would it be not suffice to use the output directory "outdir"? I am concerned that the file cache will grow whenever a java file is added or removed.
2. The way encoding is used will break on [non-ascii](https://github.com/waf-project/waf/blob/master/waflib/Task.py#L849) characters in python3
3. The file is indented with tabs, but it seems that spaces are used. Are you using Python2 or Python3? Or are you perhaps using Jython?

I can rework the changes, but I would like to have an answer for the first point.

jbendes:
Ahh sorry about the spaces issue, I have that configured in my vimrc.

For the uid, I'm actually just trying to use this waf tool more than once.
It's always on different source files, but the uid of all javac_tasks that
are created seem to be the same. I'm happy to rework the changes on the
pullreq. What would you recommend using for the uid hashing? I tried to
mirror the other tasks by using self.srcdir, since task, by default uses
self.inputs and self.outputs.

Also, I believe I'm using Python 2 but I can double check if you need me to.

Jonathan

ita1024:
The build order that you set would prevent java tasks from getting executed in parallel. The build order that you set was for non-java tasks, or?

Besides that, Java tasks with identical input folders but distinct output folders need a distinct identifier.

Please try 282f55abd3ebb8d5712ee3c22ca629d1057531a0 and report any problem.

ita1024:
I think we can close this then. Thanks for the report.

PR #1728, opened 2016-04-12
Add PlatformToolset ver. to msvs proj template

jimmyharris:
This patch adds support for an easier syntax for customizing which
version of Visual Studio msvs.py targets. It also allows users to
specify the PlatformToolset property on projects generated by waf. This
is necessary in order to properly generate projects for newer versions
of Visual Studio 2015.

Example Syntax:

To generate a solution targeted at visual studio 2012 use the following
example:

```Python
class msvs_2012(msvs.msvs_generator):
    cmd = 'msvs2012'
    numver = '13.00'
    vsver = '2012'
    platform_toolset_ver = 'v120'
```

This will populate the associated fields correctly for msvs 2013.


jimmyharris:
Crap! I just noticed something I didn't want to have happen. I will fix it locally.

PR #1729, opened 2016-04-12
Add PlatformToolset ver. to msvs proj template

jimmyharris:
This patch adds support for an easier syntax for customizing which
version of Visual Studio msvs.py targets. It also allows users to
specify the PlatformToolset property on projects generated by waf. This
is necessary in order to properly generate projects for newer versions
of Visual Studio such as 2013 and 2015.

Example Syntax:

To generate a solution targeted at visual studio 2013 use the following
example:

```Python
class msvs_2013(msvs.msvs_generator):
    cmd = 'msvs2013'
    numver = '13.00'
    vsver = '2013'
    platform_toolset_ver = 'v120'
```

This will populate the associated fields correctly for msvs 2013.

### DIFF
--- a/waflib/extras/msvs.py
+++ b/waflib/extras/msvs.py
@@ -105,7 +105,7 @@ PROJECT_TEMPLATE = r'''<?xml version="1.0" encoding="UTF-8"?>
 	<PropertyGroup Condition="'$(Configuration)|$(Platform)'=='${b.configuration}|${b.platform}'" Label="Configuration">
 		<ConfigurationType>Makefile</ConfigurationType>
 		<OutDir>${b.outdir}</OutDir>
-		<PlatformToolset>v110</PlatformToolset>
+		<PlatformToolset>${project.platform_toolset_ver}</PlatformToolset>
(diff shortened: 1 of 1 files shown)


jimmyharris:
Sorry for the re-submission. I discovered that it would be necessary to set the PlatformToolset field of the project file based on which version of visual studio was in use. This mechanism provides the ability to do that in any given file.

It would be greatly beneficial if we could get this merged upstream.

waflib/extras/msvs.py
```diff
@@ -715,6 +716,9 @@ class msvs_generator(BuildContext):
 	'''generates a visual studio 2010 solution'''
 	cmd = 'msvs'
 	fun = 'build'
+	numver = '11.00' # Visual Studio Version Number
+	vsver  = '2010'  # Visual Studio Version Year
+	platform_toolset = 'v110' # Platform Toolset Version Number
```
ita1024: You probably meant `platform_toolset_ver`. See d3b978b905580a61597b958b059646a7d9ea9334

PR #1746, opened 2016-05-24
Fixes missing backslash for REG_PYTHON_DIR

juanalday:
Some Windows python distributions (like Anaconda) add a registry entry in InstallPath without a trailing backslash (i.e C:\Python\Anaconda2 rather than C:\Python\Anaconda2**\**)
For those distributions, REG_PYTHON is invalid

- waf.bat configure
- Using C:\Python\Anaconda2python.exe
- '"C:\Python\Anaconda2python.exe"' is not recognized as an internal or external command, operable program or batch file.

This one liner adds a backslash to the directory variable if not present...


ita1024:
Maybe the fix should go to those distributions instead.

PR #1751, opened 2016-06-05
Fixes #1748 : doxygen convert */ from EXCLUDE_PATTERNS to **/ as it is used as a glob exclude filter in the code

### LINKED ISSUE #1748: doxygen: EXCLUDE_PATTERNS format incompatible
In Doxyfile exclude patterns are supposed to be written as for example:

`EXCLUDE_PATTERNS = */test/*`

But since waf uses the ant_glob for this it will not work. It works if it is written as

`EXCLUDE_PATTERNS = **/test/*`

But then the file is not really Doxygen configuration compatible. 

Also the example in the playground doesn't indeed work:

`EXCLUDE_PATTERNS       = */.*/*`

works just as

`EXCLUDE_PATTERNS       = **/.*/*`

fedepell:


fedepell:
My only doubt is if it could suffice to just change the first match instead of all (limiting the replace to the first match) but matching is for sure working this way as well.

PR #1752, opened 2016-06-05
qt5: Fix usage examples as libraries in qt5 are named QT5XXX and not QTXXX as in qt4 (ie. QT5WIDGETS instead of QTWIDGETS)

fedepell:
Just fix the documentation examples, as the code is correct and working.


ita1024:
Thanks!

PR #1754, opened 2016-06-11
Fix: when using the ut_cmd or Option --testcmd, parts of test command line may get lost

fedepell:
Fix: when using the ut_cmd or Option --testcmd, if the test command line has parameters (added for example with the ut_fun method as in the examples) this will be lost as just ut_exec[0] was used. Now join the command line before using it in substition.

### DIFF
--- a/waflib/Tools/waf_unit_test.py
+++ b/waflib/Tools/waf_unit_test.py
@@ -127,7 +127,7 @@ class utest(Task.Task):
 
 		testcmd = getattr(self.generator, 'ut_cmd', False) or getattr(Options.options, 'testcmd', False)
 		if testcmd:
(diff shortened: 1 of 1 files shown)


ita1024:
We cannot apply this in waf 1.8 because it will break existing behavior. In waf 1.9 the atribute may be removed entirely.

fedepell:
But for 1.8 is the current behavior really wanted and not a bug?

I mean currently the example test from:

 https://github.com/waf-project/waf/blob/master/demos/unit_test/tests/test1/wscript_build#L15

Will be executed as when waf build as:

`unit_test_program --help`

While if we run waf build--testcmd "valgrind %s" it will be executed as:

`valgrind unit_test_program`

and not as one would expect:

`valgrind unit_test_program --help`

The patch should fix this.

PR #1757, opened 2016-06-14
Context.recurse: ability to change wscript file name

u3shit:
Allow to load files that are not named `wscript`, useful if you want multiple wscripts to group things together but can't put them into different folders.


ita1024:
Sorry, but we will not apply contributions that include profanity. Open another pull request with a more pleasant name/account.

PR #1758, opened 2016-06-15
Fix autoconfig

guludo:
Hi!

The reconfiguration shouldn't invalidate the user's build options. For example, if the project has more than one task generator and autoconfig is set as 'clobber', then a reconfiguration triggered when the command `waf --targets=foo` is issued would make the build command post all taskgens instead of only the one named as 'foo'. This PR intends to fix that.

Best regards,
Gustavo Sousa

### DIFF
--- a/waflib/Scripting.py
+++ b/waflib/Scripting.py
@@ -593,7 +593,7 @@ def autoconfigure(execute_method):
 					Options.options.__dict__ = tmp
 			else:
 				run_command(cmd)
-				run_command(self.cmd)
+			run_command(self.cmd)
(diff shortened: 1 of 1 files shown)


ita1024:
Use the above instead.

guludo:
@ita1024 Thanks!

I think there is one error though: the build command isn't executed for 'clobber' auto configuration. I've updated this PR (force pushed) with a commit that fixes that. Could you please take a look?

ita1024:
Ouch, bad indentation. Thanks.

PR #1760, opened 2016-06-20
Extend check_swig_version

danbar:
Extend check_swig_version in order to behave similar to check_python_version.


ita1024:
If Swig is too old, then the version will always be green, never yellow.

ita1024:
Looks like copy-paste :-) Can you update the pull request with the fix above so that the correct message is displayed at least?

danbar:
You are right :-). I have updated the pull request.

ita1024:
I want to apply the changes to the branch 1.8 as well, is there any chance to get this as a single commit? Else I will have to merge by hand :-(

ita1024:
Looks like I will have to merge this by hand

danbar:
Sorry, I have read your comment a bit too late.

PR #1764, opened 2016-06-27
Fallback to run_regular_process if we can't pickle arguments

guludo:
Not sure if that's the best approach, but it's one that looks okay from my point of view.


ita1024:
Thanks!

PR #1775, opened 2016-07-25
Update cppcheck.py to work with newer waf APIs.

jimmyharris:
The cppcheck.py tool appears to not have been updated since changes in
waf 1.8 that changed the nature of build environments.  This patch
modifies the cppcheck tool to correctly determine options during the
build context and correctly generate commandlines when the env.CPPCHECK
variable is a single item list instead of a string.


jimmyharris:
I am not entirely sure if this is compatible with 1.9.x (My company uses 1.8.x internally). If there are any glaring issues please let me know.

ita1024:
`env.CPPCHECK` is supposed to be a list.

The following will work in waf 1.9 9332a9c24dd247550e2a19098754a8efb63364dd

ita1024:
I would like to know how and why `env.CPPCHECK` could become a list of lists of string objects. Please make sure that the pull request is meant for the right branch: master is currently waf 1.9 but there is a branch called waf-1.8.

jimmyharris:
Ok, these are definitely some things that have changed since 1.8.2 I will add appropriate patches to the waf 1.8 branch.

ita1024:
Consider upgrading to Waf 1.9.

PR #1782, opened 2016-07-28
Support flexflags in TaskGen

### LINKED ISSUE #1781: Flex taskgens do not support flexflags
Need support for flexflags to provide source specific parameters ie. -P[prefix]

Whyki:
Add support for flexflags in taskgens using '.l' files.  Fixes #1781


ita1024:
Ok for the idea, but we cannot apply this as there is no rebuild when the flags change.

PR #1794, opened 2016-08-14
Add configuration option to select PySide2 Python QT5 bindings, while PyQt5 bindings remain otherwise searched first by default.

fedepell:


ita1024:
Are you certain that the behavior is unchanged? When the pyqt5 options are not loaded, the getattr lines default to `want_pyside=True`.

PR #1802, opened 2016-08-19
Add suppport for finding lib boost_python in gentoo

syntion:
* find libname pattern "libboost_python-2.7.so"

Change-Id: I19a66a5992f7d077c2ed8e35e26e414ab194a7e6


ita1024:
Ouch my eyes :-)

PR #1808, opened 2016-08-25
Fix error when adding new items to SAVED_ATTRS

guludo:
If something was added to SAVED_ATTRS, the next build command failed because
the saved dictionary in the pickle file didn't have that new key.

### DIFF
--- a/waflib/Build.py
+++ b/waflib/Build.py
@@ -316,7 +316,7 @@ class BuildContext(Context.Context):
 					Logs.debug('build: Could not pickle the build cache %s: %r', dbfn, e)
 				else:
 					for x in SAVED_ATTRS:
-						setattr(self, x, data[x])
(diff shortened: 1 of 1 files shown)


ita1024:
This will leave errors unnoticed. Changing the data structures should be followed by a distclean.

guludo:
The docs say that the attribute should be a dictionary, so having an empty
dictionary as default should fine. After doing a distclean, the behavior
would be basically the same, no? I mean the default values would be empty
dictionaries.

What kind of errors do you refer to?

Anyway, if this isn't going in, I believe a message to the user asking him
to do a distclean would be nice, because it won't give the wrong impression
that something unexpected occur and also will save him some time trying to
understand what is going on.

guludo:
Oh, you've just merged, hehe
Thanks :-)

PR #1817, opened 2016-09-25
Xcode6 updates

mimon:
Fixes a bug where XCode would crash when generated by xcode6 tool combined with the c_config tool. The xcode6 merges the entire conf.env into XCode and the crash occurred if any value of the conf.env was a dictionary (XCode does not allow dictionary types in the build settings tree).


ita1024:
This will fail on Python 3 (`RuntimeError: dictionary changed size during iteration`)

ita1024:
The reason for removing `conf.env.ARCHS = 'x86_64'` from the example is unclear. Do you mean this should never be set or that this should be undocumented?

mimon:
argh, i was afraid of that .. let me fix it! Do you want a new PR?

"The reason for removing conf.env.ARCHS = 'x86_64' from the example is unclear. Do you mean this should never be set or that this should be undocumented?"
Sorry, it was a bit unclear yes. I removed it because I thought I'd be better not to limit the generated XCode project example to x86_64. But, at second thought, it might be better to that at a later time when I have a complete iOS / ARM examples ready (which is in my TODO).

mimon:
Closing and making new PR.

ita1024:
Squash your commits and update the PR. Removing ARCHS from the example looks incorrect, if this has an effect, comment the line out and add a comment in the code that explains the valid values and the effect.

PR #1818, opened 2016-09-25
bugfix: do not merge env variables of type dictionary in xcode6 tool

mimon:
Fixes a bug where XCode would crash when generated by xcode6 tool combined with the c_config tool. The xcode6 merges the entire conf.env into XCode and the crash occurred if any value of the conf.env was a dictionary (XCode does not allow dictionary types in the build settings tree).


ita1024:
Dict comprehensions were introduced in Python 2.7, so the change fail on Python 2.5 with a `SyntaxError`.

mimon:
Oh, my bad.

So is something like this more suitable?

invalids =[]
for k,v in dictionary:
  if isinstance(v, dict):
    invalids.append(k)

for k in invalids:
  del dictinonary[k]

?

ita1024:
Is that really necessary?

```python
for k, v in list(dct.items()):
    if isinstance(v, dict):
        del dct[k]
```

Are lists and sets also permitted?

mimon:
Yeh, no that's more compact and elegant. Lists are permitted. Sets are not, so I should truncate those as well!

ita1024:
What about lists of dicts?

mimon:
Well, if that's likely to appear in the .env, then yes, we have to recursively remove all invalid values.

ita1024:
We are not adding such objects there, but it would be a good thing to only allow strings and lists of string. I do not think recursion would make any sense.

ita1024:
Any news about this, or do you want me to write it?

mimon:
Sorry for the delay. Feel free to write it. Remember to return dct. I'm at work right now and can finish this later in a couple of hours otherwise.

ita1024:
I can wait

mimon:
I have updated the commit. I also added some small playground/example code and some docs sentences as well. Hope it's OK!

ita1024:
Fine, thanks!

PR #1819, opened 2016-09-26
Less verbose output when a task failed if not running with -v

jhasse:
Currently waf prints the task and the command when a build task failed. Especially with linker errors this results in a lot of output. For example when linking hundreds of object files `str(self)` lists all of them as inputs and `last_cmd` lists all of them again as they are passed to the linker.

This PR changes the behavior so that the verbose output is only shown when running with at least `-v`.

### DIFF
--- a/waflib/Task.py
+++ b/waflib/Task.py
@@ -410,7 +410,10 @@ class TaskBase(evil):
 
 		:rtype: string
 		"""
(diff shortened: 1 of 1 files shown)


ita1024:
The outputs could even be made a little bit shorter

PR #1823, opened 2016-09-30
Don't fail on output decoding errors

jhasse:
Sometimes my tasks have invalid characters in their output (for example a source file which is encoded in an encoding different from `sys.stdout.encoding` or a test program which has a bug). Currently waf fails with an exception without showing me the output at all.

This PR uses `errors='replace'` which will replace any decoding errors with ? characters, just as most terminals would do anyway. AFAIK it's only available since Python 2.7, that's why I've added the try/except.

I've tested this with Python 3.5 on Windows and on CentOS 6 with Python 2.6.


jhasse:
Thanks :)

PR #1829, opened 2016-10-07
Allow using directories as waf tools

mortenvp:
Sometimes it is useful to be able to add a module to waf as a tool.

Using this patch one can use `./waf-light configure build --tools /tmp/mytool`

This will add the files under /tmp/mytool under /waflib/extras/mytool. Such
that they can be imported in a wscript as `from waflib.extras import mytool`.

### DIFF
--- a/wscript
+++ b/wscript
@@ -232,11 +232,28 @@ def create_waf(self, *k, **kw):
 	if zipType not in zip_types:
 		zipType = zip_types[0]
 
-
+	directoryFiles = {}
 	files = []
 	add3rdparty = []
 	for x in Options.options.add3rdparty.split(','):
-		if os.path.isabs(x):
+		if os.path.isdir(x):
+			# Create mapping over files in directory to location in
+			# waflib.extras
+			root_dir = os.path.basename(os.path.normpath(x))
+			
+			for root, _, file_list in os.walk(x):
+				for file_name in file_list:
+					relative_dir = os.path.relpath(root, x)
+					relative_file = os.path.normpath(
+						os.path.join(root_dir, relative_dir, file_name))
+						
+					file_from = os.path.abspath(os.path.join(root, file_name))
+					file_to = relative_file
+					
+					directoryFiles[file_from] = file_to
+					files.append(file_from)
+			
+		elif os.path.isabs(x):
 			files.append(x)
 		else:
 			add3rdparty.append(x + '.py')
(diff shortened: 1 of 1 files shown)


ita1024:
This is interesting, but the functionality is not very well defined and there is a risk that the functionality will not last for long:
 * The comment "Create mapping over files in directory to location in waflib extras" is confusing me. Can you describe how the folders are packed in the resulting archive: do they end up at the root or under `waflib/`? Are you trying to create hierarchies of the form `waflib/extras/level1/level2/level3` ?
 * Can you explain how you came up with the name `directoryFiles`?
 * Could you use `ant_glob` in order to avoid packing unnecessary temporary files and folders (`.git`, etc)?

mortenvp:
Hi @ita1024

Yes, the mapping idea was basically an attempt to re-use as much of the existing functionality as possible.

What it does it to create a mapping between it's absolute path to it's relative path in some module folder e.g.

```
{"/home/user/dev/mylib/__init__.py": "mylib/__init__.py",
 "/home/user/dev/mylib/lib.py": "mylib/lib.py",
"/home/user/dev/mylib/sub/sub.py": "mylib/sub/lib.py"
}
```

The keys of the dict is added to the `files` list, and finally we look in the dict when creating the `tar`. Such that the files end up under the `waflib/extra/mylib/__init__.py` etc.

Now, that I'm looking at this again. I guess I should also take care for the zip file. That is probably broken currently since I only focused on the tar archive.

The name `directoryFiles` may not be a good name? Maybe `directoryMapping` or something similar would be better? I can also add more comments in the code if you like - I just followed the current style :)

Good point about filtering out temporary files etc. I guess only relevant will be `.py` files since this is really about adding a python library / module to the waflib.

ita1024:
I find the following to be confusing:
  * The comment `Create mapping over files in directory to location in waflib extras` does not explain any of the above. Some instruction for testing in the comments would be adequate.
  * The name `directoryFiles` is only marginally better than `data` (the current style does not seem to feature any variable name in `camelCase` by the way)
  * The implementation adds suspicious data shuffling such as `file_to = relative_file`

It is good that the files included do not clash with system modules, but there should be at least some rule to exclude .pyc/.pyo files from the archive, and the zip archives must contain the same files as the tar archives.

Could you update your pull request please?

mortenvp:
Hi @ita1024, 
Thanks for the feedback. I will try to update the pull-request accordingly.

One question, that came up while doing the changes: 
The location in the zip files, looks like it is missing some nesting:

https://github.com/waf-project/waf/blob/master/wscript#L292

Should it not be:
```
return os.path.join('waflib','extras', os.path.basename(x))
```

mortenvp:
@ita1024 I've updated the pull-request. Hope it is better now? I've also updated the location in the zip. Let me know if that was a mistake.

PR #1832, opened 2016-10-12
Playground sample on how to chain both qt5 tool and pyqt5 extra to use both at the same time in a waf project

fedepell:
Time ago we discussed about qt5 and pyqt5 having to be used at the same time in a waf project. As they both override some extensions this is not possible natively.

As in my project I needed to use the two together I did a small extra loaded locally that overcomes this problem chaining the call depending on the feature and I was thinking it may be useful for someone else and propose to put it into the playground (as having it as a real extra is probably excessive).

The sample will build the previously present qt5 demo (I had to change just a reference to a resource) and pyqt5 playground example showing how the integration is immediate.


ita1024:
why not.

PR #1840, opened 2016-10-24
qt5: pass cppcheck_skip=True flag when doing configuration test build otherwise build will fail if cppcheck extra loaded after qt5 (or will execute cppcheck during config phase which is useless anyway)

fedepell:
If both qt5 and cppcheck are used at the same time we have two cases:

1) If qt5 is loaded first then the qt5 compilation test code executed during "waf configure" will fail since cppcheck will be called, but this is not yet configured, doesn't have an executable name set and will fail:

`WafError: Execution failure: ['--inconclusive', '--report-progress', '--verbose', '--xml', '--xml-version=2', '--max-configs=[]', '--language=c++', '--std=[]', '--enable=[]', '/home/fede/waf/qt5test/build/.conf_check_ab611691f58b1fe96719f28f2ae060c2/test.cpp', '-I/usr/include/qt5/QtCore', '-I/usr/include/qt5', '-I/usr/include/qt5/QtWidgets', '-I/usr/include/qt5/QtGui']
`
2) If cppcheck is loaded first configure will work but additional cppchecks will be done on it which I don't think makes much sense (for speed, logic, possible failure of future tests and pollution with xml files).

In my PR I propose to just pass always the cppcheck_skip flag which will just disable cppchecking on the file. If no cppcheck is used this flag will still be there but it is harmless. I don't know if it is worth checking if the other extra is loaded and then passing it, would just overcomplicate it.

### DIFF
--- a/waflib/Tools/qt5.py
+++ b/waflib/Tools/qt5.py
@@ -508,7 +508,7 @@ def configure(self):
 		if flag:
 			msg += 'with %s' % flag
 		try:
-			self.check(features='qt5 cxx', use=uses, uselib_store='qt5', cxxflags=flag, fragment=frag, msg=msg)
+			self.check(features='qt5 cxx', use=uses, uselib_store='qt5', cxxflags=flag, fragment=frag, msg=msg, cppcheck_skip=True)
 		except self.errors.ConfigurationError:
 			pass
 		else:


xbreak:
Isn't this an issue in `extras/cppcheck.py` and not qt5?

fedepell:
I would say it's an interaction between the two and either qt5 tells cppcheck not to act (as proposed) or cppcheck doesn't act in such conditions. The whole point is that in qt5 a "normal build" is done to test if qt5 compiles so also cppcheck gets awaken.

I don't know if it would be better to fix it on the cppcheck side, such as in cppcheck_execute check if we are in configuration and in that case skip task creation.

ita1024:
@fedepell I agree with @xbreak, the problem is in cppcheck.

fedepell:
ita1024 is there an easy/standard way to know from a task that we are in the configure step? If so I'd say we can just skip cppcheck activities if we are in such step.

ita1024:
Configuration tests from `Configure.run_build` bind the configuration context object to the temporary build context. In other words `hasattr(bld, 'conf') == True`.

fedepell:
@ita1024 Thanks for the hint! I'll now try to fix it cppcheck side and create a new PR. 
Thanks again! 
fp

PR #1841, opened 2016-10-25
cppcheck: do not execute task when in configuration mode as this can break other tools (ie qt5) which do a test build. see also #1840

fedepell:


ita1024:
`if .. == True` :-1:

fedepell:
@ita1024 ew very sorry for that :) fixed now

PR #1842, opened 2016-10-25
cppcheck: add possibility to use -j option for multiple parallel threads, speeds up drastically when executing on multiple source files

fedepell:


ita1024:
Why not use the default jobs configuration corresponding to `waf -j`? (that value is displayed by waf --help`

fedepell:
Well with -j passed to waf it would not have any effect when a single build rule has more files. So for example

`bld.program(source='1.c 2.c 3.c 4.c 5.c', target='ex-prog-c')`

Generates a single task calling cppcheck just once on all 5 files, so more or less something like:

`cppcheck <various options> 1.c 2.c 3.c 4.c 5.c`

So it will anyway just check one file at a time and be potentially slow, and here the new proposed flag solves the problem as it goes in this invocation and makes it work in parallel.


Of course if you have multiple separate build rules then (apart for the bug reported in #1844 which prevents currently this working) passing -j to waf solves the problem, so if you have:

```
bld.program(source='1.c', target='ex-1')
bld.program(source='2.c', target='ex-2')
bld.program(source='3.c', target='ex-3')
bld.program(source='4.c', target='ex-4')
bld.program(source='5.c', target='ex-5')

```
Then passing -j to waf makes the trick and the new feature doesn't (as each single invocation of cppcheck has just one file).

So I would say the two things are complementary somehow in my opinion (in our personal development case we use more the first case as we usually have separate scripts for each artifact and use globs on a directory with all the sources).


Changing in general invocation of cppcheck in a way that itself will separate the calls (so starting from first configuration generating calls as in the second one) may be an idea but of course there are drawbacks in this, for example you generate more processes and the output then has to be recombined into a single report (which is something desiderable if you have many files)

Thanks!
F.

fedepell:
Sorry: or did you mean we could use the value passed to "-j" also as the value to pass to cppcheck? (so for example you call "waf -j 10" and I use this 10 also for cppcheck and call it "cppcheck -j 10")

So we have one single configuration.

In principle it would simplify the interface, but we one may have the need to decouple the two. Two theorical cases that come now to my mind:
1) You want to check for unusedFunctions in cppcheck (which cannot be done if -j > 1) but still have parallelism in waf (so -j to waf is > 1 but cppcheck not)
2) You want to keep waf single thread not to execute other stuff before cppcheck finished (and can make the wscript exit if configured so) but want cppcheck to be fast (so waf -j =1 and cppcheck >)

What I could do in case is use waf -j option as the default for cppcheck one. But this would change default behaviour and is therefore backward incompatible (that's why I put default to 1).

ita1024:
Ok, sure. I cannot look into this now though.

fedepell:
No hurry at all of course. I have it locally for my projects in use, then if it's useful for upstream then I'm even happier!
Thanks!

ita1024:
The condition `'unusedFunction' in conf.options.cppcheck_bin_enable or 'unusedFunction'...` makes a very long line. Who will even update that warning once cppcheck supports `-j` with `unusedFunction`?

PR #1844, opened 2016-10-26
cppcheck: fix extra forn multiple build rules are in a single wscript

fedepell:
When executed the output from cppcheck will be put inside cppcheck.xml and
then the generated error output inside cppcheck/index.html (and related
subfiles). Of course if two separate build rules are present the files
will clash with each other and data will be lost.

So this will not work in previous version:

bld.program(source=bld.path.ant_glob('src/ex-prog-_.cpp'), includes='src/', target='ex-prog-c')
bld.program(source=bld.path.ant_glob('src/ex-prog2-_.cpp'), includes='src/', target='ex-prog2-c')

In the output just one of the two results will be there (or in worst case
we will have files being deleted/garbled) as they both try to work on
cppcheck.xml and index.html (in build and build/cppcheck respectively)

With this commit the xml/html files have a reference to the task name (so
appended with a dash) so they are unique and don't clash. Also all the
messages to the user are corrected accordingly so the user is pointed to
the correct name of the file (and so are the internal links generated in
the html file).

In the previous case we will have:

ccpcheck detected (possible) problem(s) in task 'ex-prog2-c', see report for details:
    file:///home/fede/waf/cppc/build/cppcheck/index-ex-prog2-c.html

ccpcheck detected (possible) problem(s) in task 'ex-prog-c', see report for details:
    file:///home/fede/waf/cppc/build/cppcheck/index-ex-prog-c.html


ita1024:
I agree that the name clash is problematic, but this change does not make it possible to force the file name to be `index.html` as before.

fedepell:
Should I add a cppcheck command line option to force the execution as exactly the old behaviour? (so index.html and potential overwriting)
Thanks!

ita1024:
Having at least one conf.env variable to trigger the old behaviour would be sufficient as it is unlikely that a project would switch between the two modes.

fedepell:
Ok no problem I implement asap. Is it fine if default behaviour is the new one or would be better to keep old one as default even if it is broken when there is more than one rule?

fedepell:
Added option to have old behaviour as requested, hope it's fine!

ita1024:
This cannot be merged because there is a conflict. Could you rebase your changes on the master branch?

fedepell:
Rebased to master, hope it's fine now! Thanks!

PR #1847, opened 2016-10-28
Replace optparse with argparse

mortenvp:
The purpose of this pull-request is to move the option parser functionality from `optparse` to `argparse`.

The primary motivation for this is that we have built a recursive dependency resolution tool for waf. 

Using this we do not a-head of time know exactly what options will be added when resolving a dependency. However, a developer may already know this in advance. `optparse` will complain if it sees an option that it does not know, whereas `argparse` is a bit for flexible.

What do you think? Is this worthwhile pursuing? 

All the best,
Morten

### DIFF
--- a/waflib/Options.py
+++ b/waflib/Options.py
@@ -10,7 +10,7 @@ Provides default and command-line options, as well the command
 that reads the ``options`` wscript function.
 """
 
-import os, tempfile, optparse, sys, re
+import os, tempfile, argparse, sys, re, functools
 from waflib import Logs, Utils, Context, Errors
 
 options = {}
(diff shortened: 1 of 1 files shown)


ita1024:
Well, no because this will break compatibility. Having clearly defined and documented command-lines is also a plus for optparse.

mortenvp:
Hi @ita1024 ,
Just such that I fully understand, compatibility in what way?

From the user/CLI point of view? Or Python versions or something different?

ita1024:
Both.

jhasse:
Maybe something for Waf 2.0?

optparse has been deprecated since Python 3.2.

PR #1852, opened 2016-11-14
waf_unit_test: add possibility to dump the environment variables before test run

fedepell:
Often when a unit test run via waf_unit_test has problems it is not immediate to study it by hand (or for example with gdb) since when waf runs it the library path is created dynamically by waf depending on the dependencies. 
This patch adds a new parameter ut_dumpenv which, if specified, is the name of a file (created where the test is executed, that is the test cwd) that will contain all the environment used by the test execution. This can be then sourced/exported to easily run the tests in the same environment they are run by waf. It was tested also with pytest where the PYTHONPATH has the same issues.
Example usage:

`bld.program(source='test/unit_test.cpp', includes=['src/'] , use=['GTEST','ex-lib-c'], target='unit_test', features='test', ut_dumpenv='envdump1.txt')`

Will create envdump1.txt with the various variables in build directory (which will properly set LD_LIBRARY_PATH to find the dependant libraries referenced in use)

### DIFF
--- a/waflib/Tools/waf_unit_test.py
+++ b/waflib/Tools/waf_unit_test.py
@@ -31,9 +31,13 @@ the predefined callback::
 		bld(features='cxx cxxprogram test', source='main.c', target='app')
 		from waflib.Tools import waf_unit_test
 		bld.add_post_fun(waf_unit_test.summary)
+
+By passing --dump-test-runner the execution will also create in the build directory a python file named as the task name with -run.py appended
+that can be executed manually to reproduce the exact execution manually for debugging purposes.
+If the script is run with arguments these will be used instead of the command itself. This can be used to run for example a debugger or profiler on the test manually.
 """
 
-import os
+import os, sys
 from waflib.TaskGen import feature, after_method, taskgen_method
 from waflib import Utils, Task, Logs, Options
 from waflib.Tools import ccroot
@@ -151,6 +155,13 @@ class utest(Task.Task):
 
 	def exec_command(self, cmd, **kw):
 		Logs.debug('runner: %r', cmd)
+		if getattr(Options.options, 'dump_test_runner', False):
+			if not self.generator.name:
+				Logs.warn('waf_unit_test: --dump-test-runner task with no name, outputs may collide')
+			runnername=os.path.join(str(self.generator.path.get_bld()), ( "%s-run.py" % (self.generator.name or 'dump-test-runner')))
+			Utils.writef(runnername, ENVFILE_TEXT %  { 'pyexe': sys.executable, 'env': self.get_test_env(), 'cwd': self.get_cwd().abspath(), 'cmd': cmd[0], 'cmdline': cmd })
+			os.chmod(runnername, Utils.O755)
+
 		proc = Utils.subprocess.Popen(cmd, cwd=self.get_cwd().abspath(), env=self.get_test_env(),
 			stderr=Utils.subprocess.PIPE, stdout=Utils.subprocess.PIPE)
 		(stdout, stderr) = proc.communicate()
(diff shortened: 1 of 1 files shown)


fedepell:
Discussing of it with @xbreak  maybe could be done as an option instead of an attribute (easier to change on the fly) with the file name generated by taskgen name or so.

ita1024:
Why open the file in `append` mode? Are you certain that there is no problem with threads?

I am afraid that this solution is insufficient for actual troubleshooting and that it creates a burden to maintain. How about having an option to have each test dump one executable `.py` file with the exact environment and command instead?

fedepell:
Sorry for the append, will fix! The operation is per test execution, so I believe it's thread safe yep.

But the idea to create directly a .py is better yes, more portable for sure and easier to maintain, with the command at the end that can be customized (ie. prepend gdb). 

I'll work on that asap and update the PR!

Thanks!

fedepell:
I updated the PR. Now python script is generated. Is an option so it's more versatile. The script can also be run customized so more versatile also for advanced usage one can thing of.
Hope it is fine!

ita1024:
Ok for the idea, but the implementation unfortunately needs some polishing:

1. Are you really believing that `--dump-test-env` a good name for what the option does?
2. Arguments for the program should never be joined as a big string because that is not how the program is executed: there is no shell.
3. The output file must use a command of the form `status = subprocess.Popen([cmd, arg1, arg2, ...], shell=False, cwd=.., env={...}); sys.exit(status)` instead of calling `os.exec` functions. Inserting `gdb -args` at the start of the list is simple enough, but mentioning this as a comment in the generated file would be a plus. Do you want people to refer to `waf_unit_test.py` in order to understand the generated test file?
4. Files must be written with Utils.writef() to ensure that the implementation works in both Python 2 and 3 and that file descriptors are properly closed.
5. The resulting file should have executable permissions.
6. Could you rebase and squash your changes? I am not sure I want to get through every commit that you make.

fedepell:
Thanks for the feedback, sure happy to polish :)
1) Indeed it was before when a env script was dumped, now that we changed to py a bit less. I'll correct!
2) No just for the ENV variable (UT_CMDLINE variable as a support for use) they are joined in a string, to the exec they are passed as an array, so I believe it should be ok as it is!
3) Indeed I started first with subprocess.call (as we don't need to communicate with it, isn't it better than popen?) but then switched to exec as I think we don't really need to return back to Python since this is just a bootstrap script. But if you think it's better, no problem to change. 
Ok to put a comment inside the generated file on the arguments!
4) Ok I correct!
5) Ok!
6) Sure no problem let me fix a few things then when it looks fine a squash to a single commit!

Thanks!

ita1024:
2. Do not join arguments or environment variables, and do not apply environment variables to the parent process
3. Do use subprocess and do return the subprocess exit code

fedepell:
I hope I fixed all the request and did a rebase to have a single commit.

Just one note on the joining: as a helper a UT_CMDLINE is created (now just in the script environment, not parent) and this has to be joined, as an env variable cannot be a tuple. This happens if you use for example ut_run parameter (when the program executed will have also parameters passed). But now is done locally inside the generated script. This variable is useful in case you run the script with additional custom parameters and want to recall how waf would call it. If you run the script without custom parameters then it would not be needed.

Hope it's fine!
Thanks!

ita1024:
I imagine that you did this in a hurry. For your information:
1. The constant is called ENVFILE_TEXT but it contains something that has little to do with environments
2. Using the task generator "name" for a file name is an unhappy choice as the executable file name could be used instead with little risk of collision
3. The docstring for ENVFILE is inaccurate and features even errors
4. I do not understand what UT_CMDLINE is for
5. I am not convinced that the generated script should handle command-line arguments
6. The script does not specify an encoding so %()s is likely to miss special characters. Those substitutions will also miss quotes if ever present
7. The argument cmd is turned into cmdline for no particular reason

ita1024:
I suggest that you update your copy, and open another pull request if you believe that something needs amending.

fedepell:
Thanks for the feedback! OK I'll put the suggested fixes in a new PR, no problem.

For the questions pending.

The script command line are to execute it easily with additional tools. For example to start it with gdb (or gdb on a core dump) or so. Without the command line arguments the user should edit the generated python file and change the execution parameters which I believe is not very handy especially if you want first to run it with gdb, then maybe manually with valgrind or some other tool.

The UT_CMDLINE is indeed just a helper for that: now when you execute it with command line arguments you don't have anymore the original command line, as the arguments are used instead of it. Using $UT_CMDLINE you can retrieve the original execution arguments (and use it after your custom gdb or whatever call)

ita1024:
I believe that this is over-engineered, but if you really consider that, then the script should provide proper command-line help (--help), the contents should be documented as well and the code should work in both python2 and python3.

Issue #1862, opened 2016-11-22
functools.partial() not accepted as rule


gjcarneiro:
Trying partial function application gives me an error:
```python
def myfunc(arg, task):
    ...
bld(...,
        rule=partial(myfunc, "arg"))
```
```
TypeError: functools.partial(<function build.<locals>. [...] at 0x7f0578342bf8>, 'arg') is not a module, class, method, function, traceback, frame, or code object
```
Sounds like waf is checking for exact types, but instead it should check for callables:
```
callable(...)
    callable(object) -> bool
    
    Return whether the object is callable (i.e., some kind of function).
    Note that classes are callable, as are instances of classes with a
    __call__() method.
```

ita1024:
I am not sure any verification is made anywhere.

ita1024:
The parameter is meant to be either a string or a hashable function.

xbreak:
If you can, use a lambda instead of partial.

Lambdas aren't hashable by `Utils.h_fun` either, but the failure will be handled by catching the resulting `IOError` as an `EnvironmentError`.

gjcarneiro:
Oh, I see, Utils.h_fun has special code to handle functions.  I knew there had to be something handle functions especially:

```python
def h_fun(fun):
	try:
		return fun.code
	except AttributeError:
		try:
			h = inspect.getsource(fun)
		except EnvironmentError:
			h = 'nocode'
		try:
			fun.code = h
		except AttributeError:
			pass
		return h
```
In theory it should be possible to handle also `partial`, as long as the args are also hashable:
```
>>> partial(int, 2).args
(2,)
>>> partial(int, 2).func
<type 'int'>
```
If I provide a PR, will you still consider reopening?

ita1024:
The code needs to work on Python 2.5-3+, on other Python interpreters (Pypy, Jython, ipy) and reflect the contents of the object being used.

gjcarneiro:
The `functools` module is "New in version 2.5.".  The partial object attributes are [documented here](https://docs.python.org/2/library/functools.html#partial-objects).  If they're documented, we can rely on them to be available in any Python implementation.  I would like to be able to hash any partial func as long as its arguments are hashable (kwargs can be converted to a tuple of (key, value) pairs for hashing purposes).

PR #1865, opened 2016-11-25
Accept functools.partial() as a task rule #1862

gjcarneiro:
Besides hashing `functools.partial` objects, I needed to tweak `process_rule` to take it into account.  The reason is that normally a function defined inside a class becomes a bound method automatically when accessed through an instance, but not so for functools.partial.  Docs say:
>  partial objects defined in classes behave like static methods and do not transform into bound methods during instance attribute look-up.

### DIFF
--- a/waflib/Task.py
+++ b/waflib/Task.py
@@ -7,6 +7,7 @@ Tasks represent atomic operations such as processes.
 """
 
 import os, re, sys, tempfile
+from functools import partial
 from waflib import Utils, Logs, Errors
 
 # task states
@@ -291,7 +292,13 @@ class TaskBase(evil):
 			pass
 
 		try:
-			ret = self.run()
+			if isinstance(self.run, partial):
+				# Python documentation says: "partial objects defined in classes
+				# behave like static methods and do not transform into bound
+				# methods during instance attribute look-up."
+				ret = self.run(self)
+			else:
+				ret = self.run()
 		except Exception:
 			self.err_msg = Utils.ex_stack()
 			self.hasrun = EXCEPTION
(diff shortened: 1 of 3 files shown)


ita1024:
The amount of maintenance work packed into such a small amount of changes is absolutely amazing! Consider the following snippet:

```python
import functools
from waflib import Utils

def fun(arg1, arg2, task, one=1, two=2):
        print(arg1, arg2, task, one, two)

par1 = functools.partial(fun, 'arg1')
par2 = functools.partial(par1, 'arg2', one=11, two=22)

print(Utils.h_cmd(par1))
print(Utils.h_cmd(par2))
```

The code is written in a rather complicated manner so it is not immediately clear that the last call to `h_fun` should be a call to `h_cmd`.

Besides that, the new code provides nice "distracters" to obfuscate the above:
1. It is using the fact that `dict.items()` results are not sorted but are dependent on the Python version, which will trigger rebuilds in non-obvious cases
2. It adds unnecessary calls to `Utils.h_list` which is a slow function, so the changes should be made on `Utils.h_fun` in order to re-use caching
3. It changes the style of the imports line for no particular reason
4. A commit history to complicate the review; fortunately Github provides a summary of the changes and a way to squash commits to mitigate this

gjcarneiro:
Thanks for merging!  Yes, I forgot about dict.items() not being sorted.  We definitely should try to sort the items, otherwise spurious builds will result.  Do you want me to provide a (trivial) commit?

ita1024:
Your suggestions [have been taken into consideration](https://github.com/waf-project/waf/commit/906bf781ac0d3bcb1e89aa568d465039b9cf086a#commitcomment-19969392) but your changes will not be needed.

PR #1869, opened 2016-11-29
Ignore errors in shutils.rmtree

JPEWdev:
Causes shutils.rmtree() to ignore errors when attempting to clean up the
temporary configuration directory. This is necessary on Windows when
executing a program during Configuration.check() because it doesn't
appear to allow the parent directory to be removed if it recently had a
running executable file in it.

Without this change, this trivial wscript fails on Windows platforms:

    def options(opt):
        opt.load('compiler_c')

    def configure(cfg):
        cfg.load('compiler_c')
        cfg.check(execute=True, msg='Executing program')

### DIFF
--- a/waflib/Configure.py
+++ b/waflib/Configure.py
@@ -594,8 +594,26 @@ def run_build(self, *k, **kw):
 			proj['cache_run_build'] = ret
 			proj.store(os.path.join(dir, 'cache_run_build'))
 		else:
-			# Ignore errors. Windows will sometimes not allow the directory to be deleted
-			shutil.rmtree(dir, ignore_errors=True)
+			def handle_error(fun, path, exc):
+				# Try the operation a few more times after sleeping to allow
+				# other processes to close open handles that might be
+				# preventing the delete from succeeding. This happens
+				# frequently on Windows due to the Application Experience
+				# Lookup Service
+				tries = 0
+
+				while True:
+					time.sleep(0.01)
+					try:
+						fun(path)
+					except:
+						if tries >= 100:
+							raise
+						tries += 1
+					else:
+						break
+
+			shutil.rmtree(dir, onerror=handle_error)
 	return ret
 
 @conf
--- a/waflib/Configure.py
+++ b/waflib/Configure.py
@@ -594,7 +594,8 @@ def run_build(self, *k, **kw):
 			proj['cache_run_build'] = ret
 			proj.store(os.path.join(dir, 'cache_run_build'))
 		else:
-			shutil.rmtree(dir)
+			# Ignore errors. Windows will sometimes not allow the directory to be deleted
+			shutil.rmtree(dir, ignore_errors=True)
 	return ret
 
 @conf


ita1024:
Tell us, which process is preventing the folder to be removed?

JPEWdev:
I don't actually know if it a specific process. The sequence of events that causes this is rather strange:

We execute the test program (testprog.exe) in a preforked process, Popen.communicate() returns and the process exits cleanly. Then, when we do the recursive deletion of the directories in the shutils.rmtree(), the os.remove() of testprog.exe succeeds (which is very strange to me), but the removal of the testbuild directory fails, claiming that the directory is not empty (WindowsError: [Error 145] The directory is not empty). If I do an os.listdir() in the code when the failure happens, it clearly shows that testprog.exe is still in the directory, but once the script exits it is obviously empty, as I can manually browse to the directory and see the file is not there. Also strange is that any attempt I make to os.remove() or os.stat() testprog.exe after the directory not empty failure results in a "Access Denied" error (WindowsError: [Error 5] Access is denied).

ita1024:
Have you tried configuring your anti-virus?

JPEWdev:
I think it is configured to ignore the directory I'm building out of, but I will double check

JPEWdev:
Alright, I tracked this down using the Process Monitor.

Turns out, Windows has a service called the Application Experience Lookup Service (AeLookupSvc) [https://support.microsoft.com/en-us/kb/902196]. This service opens the test executable to try and do something with it (not sure exactly what), but doesn't end up closing the file until after waf has attempted to delete the directory, causing the error to occur.

Annoyingly, this service is enabled by default from Windows 2003 Server and later (I'm running Windows 7). I think the fundamental issue (other than that this service is being annoying) has two factors:
 1. The test programs that waf executes don't run for a very long time
 2. Waf tries to delete the directory containing the executable immediately after it finishes executing.

Both of these combine to cause the Application Experience Lookup Service to be unable to close the file handle fast enough for the shutils.rmtree() to succeed. Since this is a default behavior of Windows, I feel like we should be handled in Waf (even though I agree it is not Waf is not at fault). How would you feel about making Waf wait for a short period of time in the event that a failure occurs during shutils.rmtree()?

```python
	def handle_error(fun, path, exc):
		# Windows systems have some internal services that like to open
		# executables when they are run and examine them (i.e. the
		# Application Experience Lookup Service). Normally, this isn't
		# an issue because the service closes the executable promptly.
		# However, here we are trying to execute a test executable
		# which will run for a very short period of time, then
		# attempting to immediately delete the directory that contains
		# that executable.  Both of these steps can happen before the
		# Windows services can release their handle on the executable,
		# which will give us an error saying the directory is not
		# empty, so we can't delete it.  All we can do is wait for a
		# short time to give the services a chance to close their handles
		# and try again
		time.sleep(1)
		fun(path)

	shutil.rmtree(dir, onerror=handle_error)

```

jhasse:
Removing directories on Windows randomly fails. Yes. I've ran into this problem with Jenkins and the only solution I found is to retry multiple times (similar to what JPEWdev suggested above).

This is how I do it:

```python
def rmdir(directory):
    tries = 0
    while True:
        try:
            shutil.rmtree(directory)
            return
        except (PermissionError, OSError) as err:
            tries += 1
            if tries > 100:
                raise err
            else:
                time.sleep(0.01)
```

JPEWdev:
Alright, I updated to the commit to try a few times to delete the file, and if it continues to fail raise an exception

ita1024:
Why should we have a fix here and not for instance:

1. In Windows: are you using the latest version from Microsoft? Have you made a bug report (link)? I am not seeing any problem on Windows 10.
2. In Python: in case this is a known OS quirk, every Python application will need a routine to wait before deleting files. Why should it not go in shutil.rmtree instead?
3. Elsewhere: there are many more uses of `shutil.rmtree` in our codebase, so why should a fix be made only for these specific configuration tests?
4. In your systems configurations: if you cannot configure your development environment, maybe experts should do it for you instead? Is that not what operations are for?

JPEWdev:
Yes, I agree that fixing this here is probably not the ideal situation and if that is really the consensus, I'm fine with it. However, I would like to have some more discussion first and make a few points:

> In Windows: are you using the latest version from Microsoft? Have you made a bug report (link)? I am not seeing any problem on Windows 10.

Running Windows 10 is definitely not a option for us or many other users. I have not made a bug report, but I can if you really think it would help. I did some research and it appears they have removed the Application Experience Service from Windows 10. I don't have a Windows 10 PC available to verify this for sure right now, but my searches seem to suggest that is the case.

> In Python: in case this is a known OS quirk, every Python application will need a routine to wait before deleting files. Why should it not go in shutil.rmtree instead?

I can try that also, but the conditions for this to occur are so specific (i.e. you have to try to delete a directory that just had a very short running executable in it) I'm doubtful the Python community would approve.

>Elsewhere: there are many more uses of shutil.rmtree in our codebase, so why should a fix be made only for these specific configuration tests?

Can we make a common Utils.rmtree() perhaps?

> In your systems configurations: if you cannot configure your development environment, maybe experts should do it for you instead? Is that not what operations are for?

I can disable the Application Experience Service and everything works without modification via `sc config AeLookupSvc start= disabled`, however you'll have a hard time convincing me that all users are required to modify the *default* behavior of their operating system to get Waf to work correctly.

Any thoughts?

ita1024:
1. The argument against using Windows 10 is ludicrous. The very first thing any serious Windows developer does is configure his system properly. At a very minimum, a ticket should be opened to confirm the Windows version(s) affected and the solutions.
2. Every software project out there will probably need such retry logic for Antivirus scanners too, so the re-try logic should not be duplicated in every Python project.
3. Utils.rmtree is an idea, but changes will not be applied unless you show us the tickets for the two points above.

JPEWdev:
>The argument against using Windows 10 is ludicrous.

The IT department at our company won't let us switch to Windows 10. We are stuck at Windows 7. Sorry.

>The very first thing any serious Windows developer does is configure his system properly.

My life would probably be a lot easier if everyone was a serious Windows developer :). Configuring my system isn't the problem, it's telling everyone else they have to run some strange command line that changes their system configuration so that than can successfully run `waf configure`. I honestly don't know what all the Application Experience Service does.... it seems innocuous enough to disable it, but I can't be sure it isn't necessary for something important.

> Every software project out there will probably need such retry logic for Antivirus scanners too, so the re-try logic should not be duplicated in every Python project.

Makes sense

I'll look in to creating some tickets at some point.

ita1024:
Just open the two tickets above. No change will be made until then.

ita1024:
Any update?

ita1024:
No update, re-open when done.

PR #1876, opened 2016-12-13
correct static lib extension for emscripten

montonero:
When linking with static libraries emscripten would consider .a files not .bc


ita1024:
@zougloub please check

PR #1878, opened 2016-12-14
Tools/python: add option --nopycache

ricardocrudo:
This might be useful to force the installation of python objects to site-packages, i.e.: not in `__pycache__` directory. I've found this need when using buildroot with the option ".pyc compiled sources only" selected. Apparently buildroot do nothing for objects placed at `__pycache__` directory.

### DIFF
--- a/waflib/Tools/python.py
+++ b/waflib/Tools/python.py
@@ -103,7 +103,7 @@ def process_py(self, node):
 		pyd = node.abspath()
 
 	for ext in lst:
-		if self.env.PYTAG:
+		if self.env.PYTAG and not self.env.NOPYCACHE:
 			# __pycache__ installation for python 3.2 - PEP 3147
 			name = node.name[:-3]
 			pyobj = node.parent.get_bld().make_node('__pycache__').make_node("%s.%s.%s" % (name, self.env.PYTAG, ext))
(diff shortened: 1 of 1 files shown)


ita1024:
Your change seems to skip .pyc processing on python3 under a certain condition, but the reason is unclear. If compiling those files in __pycache_ does nothing, then what is the problem? Are .pyc files unnecessary with buildroot or are they causing problems? And what about python2?

ricardocrudo:
The change force to install .pyc to site-packages directory (as in python2) instead of pycache, but only if the flag `--nopycache` is passed.

The buildroot has the option ".pyc compiled sources only" which means that only .pyc files are kept in the final deploy, in other words all .py files are removed. This is an option to save space and, I guess, make it faster. The issue is that one can't load modules from pycache directory. Here is a quote from python3 modules [docs](https://docs.python.org/3/tutorial/modules.html#compiled-python-files):

> To support a non-source (compiled only) distribution, the compiled module must be in the source directory, and there must not be a source module.

ita1024:
In this case, should the build skip the `.py` file installation too? Or does buildroot exclude those?

ricardocrudo:
buildroot does that.

ita1024:
I can add this now, but the option really needs a better name to reflect its purpose.

PR #1879, opened 2016-12-14
glib2: consider destdir when compiling schemas

dffischer:
When a `destdir` is set, the gschemas are already installed below it, but the compilation process is still invoked at the path directly below the root directory. The change in this pull request directs it to compile the newly installed files.


ita1024:
If `env.GLIB_COMPILE_SCHEMAS` contains a list (a command with spaces) then `env.get_flat('GLIB_COMPILE_SCHEMAS')` will return space-separated list of arguments. The command will thus fail to run.

I also think that using a simple if-else block would make the code a bit more obvious.

dffischer:
Switching back to a command string instead of a tuple should also support lists. I also replaced the inline `if` with a block.

ita1024:
Unfortunately this still introduces a problem if a command contains a space:
`['/path/to/a folder/space']` -> `$ /path/to/a folder/space`

I guess I will have to fix this myself.

dffischer:
I was unsure how to correctly use environment variables in `exec_command`. Thank you for the merge and fix.

Issue #1699, opened 2016-02-09
Boost library name templates used in search do not follow the boost library naming convention


yunuscanemre:

ita1024:
It is probably because it was written a while ago. Patches welcome.

yunuscanemre:
alright, thanks. updated the details for the issue.

johan-boule:
I'm not sure if you're talking about a specific compiler. Be careful not to break gcc/mingw, clang or whatever. Make your special case also depend on the *compiler*, not just the operating system.

PR #1880, opened 2016-12-17
Fix boost library naming conventions

cconverse711:
Uses proper filename conventions, allows static ABI to be specified, as requested per #1699


ita1024:
Thank you

PR #1881, opened 2016-12-21
glib2: Compile schemas per directory

dffischer:
By changing `GSETTINGSSCHEMADIR` during the build setup or on single tasks or generators, the user may place schemas in various locations. Adding a post build function for each of this location compiles all of them instead of only one global directory.

For example, extensions to the gnome-shell often store their schemas in a local `schemas` subdirectory. When using waf to build them, this patch allows installing multiple in one go.

I chose to use `functools.partial` here as it allows reuse of most of the existing code. If deemed more appropriate, I could rewrite `compile_schemas_callback` to return a function instead.


ita1024:
The implementation is unfortunately incomplete:
1. An example is needed in demos/glib2
2. Error handling is unspecified: what should happen when one of the updates fail? Is it OK to ignore all errors? If yes, then that should be mentioned.
3. Only one callback function is required. Since data is bound to the `bld` object, a new callback function can be added to loop and call `compile_schemas_callback` as many times as needed. Using `functools.partial` only makes the code much more complex than necessary.

dffischer:
Thank you for the quick feedback. I changed the callback to run once and process a set of directories.

Concerning the error handling, one of the compilations may indeed fail, for example when the output file already exists and is set immutable. Compilation of other files should then still continue. I added a warning in this case.

As an example was already in place at `demos/glb2`, I augmented it with the new functionality. Moving it to a subdirectory and creating a dedicated one beside it would also be possible, similar to the multiple examples in the c or vala demos.

I hope my code does not contain too much Python2 inadeptness, I usually only deal with Python3.

dffischer:
Thank you for the merge and, as always, the quick reaction.

PR #1888, opened 2017-01-05
Add qt5 5.8.x new libraries to qt5 tool

fedepell:


ita1024:
I wonder how many more will be added

fedepell:
Heh, every release they add a few :)

One question that was crossing my mind: do you think it would make sense to add an option to the QT5 tool to automatically generate (or just automatically add the missing ones) the list of libraries from the system installation?
This could be useful especially when you have additional QT5 modules (that are not in the standard QT5 base distribution but were installed later) on the system otherwise you have then to manually add them. 

Or add an option to add specifically elements from command line to this QT5_LIBS (ie. -qt5-add-lib=Qt5OpcUA will add this to the list). Well one could create a qt5_vars complete list...

If it sounds that it makes sense I can prepare a PR.

ita1024:
Yes.

PR #1891, opened 2017-01-11
qt5: add --qtlibsfromsys option to generate the list of Qt5 libraries by searching on the system

fedepell:
This option gives the possibility to search on the system (QT5_LIBDIR or library automatically
found) for libraries available instead of using the list hardcoded in the tool. The search is
done using a regexp that matches the same files as the ones used for library search with support
for dynamic/static and win32/unix.

This option can be useful in a couple of situations:
-) A Qt5 version that is not (yet) supported by waf and therefore waf doesn't have the libraries yet
-) A custom Qt5 version with many unofficial libraries (that otherwise have to be added by hand
or using --qtextralibs)
-) A very small Qt5 installation (ie. embedded systems) where otherwise waf configure will anyway check
for the big list of libraries (and not find most of them)

### DIFF
--- a/waflib/Tools/qt5.py
+++ b/waflib/Tools/qt5.py
@@ -70,7 +70,7 @@ except ImportError:
 else:
 	has_xml = True
 
-import os, sys
+import os, sys, re
 from waflib.Tools import cxx
 from waflib import Task, Utils, Options, Errors, Context
 from waflib.TaskGen import feature, after_method, extension, before_method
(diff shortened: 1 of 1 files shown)


ita1024:
I disagree with this design. Adding a new option means that the code will never be tested properly and the qt5 file is already too complex. By the way, `filter` does not return a list in Python3, so the call to `sort()` is likely to fail.

I suggest to replace or to complete the usage of `QT_LIBS` with a dynamic listing of the Qt5 libraries in the qtlibs folder. Re-using `conf.env.cxxstlib_PATTERN` and `conf.env.cxxshlib_PATTERN` could be a good idea.

fedepell:
Sorry for the delay, have a very busy moment!

Ok I'll check the cxxstlib_PATTERN code and try to make a better proposal, no problem! (if it is annoying to have the PR open for now just close and I'll reopen)

As for using this as a default without option: I think it would be a good idea. Anyway the missing libraries and caught later on at build execution so it wouldn't change anything from a practical perspective I believe. And it would ease maintenance for future versions of QT or installation with non default modules.

fedepell:
Hi!
I did some changes hopefully as requested:
1) This method is now the default, no additional option. This also cleans up the big list of libraries in QT5_LIBS to be maintained with new releases
2) Use lib_PATTERN to have the library pattern and then customize to Qt5.
3) Use sorted() instead of .sort() for Python compatibility

I tested it on a few Linux and Win systems with dynamic and static libraries and looks to work fine!

If you think it could work I will rebase to master and squash into a single commit.

ita1024:
This is better. Note that the following will need to be reworked:
  - The condition `if qtliblist` will always match in Python 3 (filter objects).
  - Regarding `qtlibs`, data is usually bound as `env.VALUE` not to `conf.value`. The former is preferable for data, while the latter is for functions/methods.

fedepell:
I should have fixed the two points hopefully! I tried with Python3 as well this time :) 
I rebased and squashed the changes, hope it's fine.

ita1024:
Yes it is. Thank you!

ita1024:
I had to make minor changes to fix the detection on windows systems. You may want to update your copy. Also, i think we can drop the debug library detection (in the branch `waf-2.0`).

fedepell:
Thanks for notifying! No problem I align now.

For debug: I'd say either drop or put it as an option?

Issue #1870, opened 2016-11-30
Autoconfigure when waf binary updates


jhasse:
When I update Waf to a new version and run `waf` I get the error

```
Version mismatch! reconfigure the project
```

even if I set `Configure.autoconfig = True`. Also "reconfigure" should be capitalized.

ita1024:
This works exactly as it should, changing the waf version must not silently delete old build information.

jhasse:
I see. Do you agree though, that the error message could explain it a little better? E. g.

```
Project was configured with a different version of Waf. Please reconfigure it using
`waf configure` (this will discard old build information).
```

PR #1893, opened 2017-01-16
Improve error message about version mismatch

jhasse:
See #1870.


ita1024:
Not all projects are configured with `waf configure` as variants and command-line options may be required.

ita1024:
Will you fix the message or are you insisting on this?

ita1024:
We cannot add a misleading message, so this will be closed. Reopen only if you fix your changes.

PR #1896, opened 2017-01-17
Respect PKG_CONFIG environment variable.

illiminable:
Currently will always search the build machines PATH and does
not allow the path to pkg-config to be pointed to an alternate
location eg. SYSROOT of target.

This matches behavior of autoconf.

### DIFF
--- a/waflib/Tools/c_config.py
+++ b/waflib/Tools/c_config.py
@@ -234,7 +234,10 @@ def validate_cfg(self, kw):
 	"""
 	if not 'path' in kw:
 		if not self.env.PKGCONFIG:
-			self.find_program('pkg-config', var='PKGCONFIG')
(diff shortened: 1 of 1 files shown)


illiminable:
This is in the context of making the Samba build handle this case - https://lists.samba.org/archive/samba-technical/2017-January/118079.html

ita1024:
The existing code uses `os.environ('PKGCONFIG')`, not `os.environ('PKG_CONFIG')`.

ita1024:
Can you use the waf-endorsed environment variable `PKGCONFIG` or do you have a problem with it?

illiminable:
We can assign PKG_CONFIG to PKGCONFIG and get Samba updated to the latest version of that function so that it works.

Thanks.

PR #1899, opened 2017-01-24
Allow numbers for Java compat attribute

DigitalDan05:
If the compat attribute was an integer the build would throw an exception due to it not being iterable.  Convert it to a string to be more flexible.


ita1024:
I think that you should submit this as an enhancement for the subprocess module on the Python bug tracker and read what they have to say about it.

PR #1903, opened 2017-02-05
Retain original trace-back

Tiksagol:


ita1024:
Why? The functions that follow might catch exceptions too.

ita1024:
It seems you are running python 2.

PR #1904, opened 2017-02-08
Update check_endianness() example to show that the return value needs…

selsky:
… to be used


ita1024:
Have you tried running the example after changing it? Where is `ctx` defined?

selsky:
Sorry for that brain fart.  I've fixed my code and verified locally.

Issue #1905, opened 2017-02-08
struct field check fails on linux


selsky:
My wscript has:
```
#! /usr/bin/env python

def options(opt):
    opt.load('compiler_c')

def configure(conf):
    conf.load('compiler_c')

    conf.check_cc(type_name="struct ntptimeval", field_name="tai",
                  header_name=["sys/time.h", "sys/timex.h"])
```
I'm running waf configure using waf from git head
[config.log.txt](https://github.com/waf-project/waf/files/759658/config.log.txt)

```
#include <sys/time.h>
#include <sys/timex.h>

int main(int argc, char **argv) {
        char *off;
        (void)argc; (void)argv;
        off = (char*) &((struct ntptimeval*)0)->tai;
        return (size_t) off < sizeof(struct ntptimeval);
}

<==
[1/2] Compiling build/.conf_check_a7c44a2d34dec0a389be34660cd5d406/test.c

['/usr/bin/gcc', '../test.c', '-c', '-o/tmp/build/.conf_check_a7c44a2d34dec0a389be34660cd5d406/testbuild/test.c.1.o']
err: ../test.c: In function ‘main’:
../test.c:8:10: error: ‘size_t’ undeclared (first use in this function)
  return (size_t) off < sizeof(struct ntptimeval);
          ^~~~~~
../test.c:8:10: note: each undeclared identifier is reported only once for each function it appears in
../test.c:8:18: error: expected ‘;’ before ‘off’
  return (size_t) off < sizeof(struct ntptimeval);
                  ^~~
```
Looks like adding stddef.h would fix this since size_t is defined there per http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/stddef.h.html

OSX and NetBSD both build without any changes.  But I get this error on Debian sid.

ita1024:
We will not make changes because the feature is deprecated and removed in waf-2.0.

PR #1906, opened 2017-02-08
Include stddef.h in SNIP_FIELD

selsky:
Make sure that size_t is defined per GitHub issue #1905


ita1024:
No, we cannot do that.

If you want a header for all configuration tests, first make sure it is actually present and has `conf.check(header_name=['stddef.h'], auto_add_header_name=True)` (or set the list `conf.env.include_key')

selsky:
Since size_t is used in the snip itself, shouldn't the snip be responsible for including any required headers?  Or should the snip avoid using size_t since some platforms don't define size_t unless you explicitly include stddef.h or sys/types.h?

ita1024:
Those headers may be missing from certain platforms, so it is up to the project to set a configuration test.

selsky:
The compiler error isn't from the headers that I'm testing though.  It's from the snippet's code.2

Can we replace the last line of the snippet with "return 0"?  Or is it critical to check that off < sizeof()?

ita1024:
Yes, you are right in the sense that the test is most likely going to only succeed on some platforms. If we make changes then we need to turn this into a chain of tests. I believe that the test code is just too complicated and the `type_name`/`field_name` should never be there in the first place (will be removed in waf 2.0).

No we cannot replace the last line: the compiler might optimize the code and render the test ineffective.

JPEWdev:
So what's the recommended migration strategy if you have a project that was using one of these checks (I'm particularly interested in function_name)?

ita1024:
@JPEWdev pass the test code as `fragment=` parameter:

https://github.com/selsky/waf/blob/8deaa2bac0c653be8c5406cab969bc03ec8c02de/waflib/Tools/c_config.py#L55

Issue #1910, opened 2017-02-15
qt5 broken on macOS


cawka:
Commit 823ad57b89b1b450b5c7b69323a2df5b5a33a230 effectively broke QT5 detection on macOS plaftorm for some QT5 installations.

It is already challenging to make compilation on macOS, as the distributed open-source version only ships with Qt*.framework and don't provide .pc files.  HomeBrew version made a few fixes, compiling .framework's but also providing .pc files.  So, this combination worked prior 823ad57b89b1b450b5c7b69323a2df5b5a33a230, but not since.

I tried to make a trivial fix by adjusting `set_qt5_libs_to_check`, but it is not as trivial, as frameworks for some reason don't include "5" in their names (e.g., `QtWebView.framework`, not `Qt5WebView.framework`), which breaks a bit of logic in this function.  I can probably make it work with a bit more cludges.

* * *
I want to ask for advice.  What is the expectation on how QT5-dependent project should be compiled on macOS?  Should we invent a way to deal with `.framework`?

cawka:
This is a quick patch that fixed my immediate problem

```patch
diff --git a/waflib/Tools/qt5.py b/waflib/Tools/qt5.py
index 6bc2370..da647a2 100644
--- a/waflib/Tools/qt5.py
+++ b/waflib/Tools/qt5.py
@@ -722,11 +722,13 @@ def set_qt5_libs_to_check(self):
                        pat = pat.replace('.dll', '.lib')
                if self.environ.get('QT5_FORCE_STATIC'):
                        pat = self.env.cxxstlib_PATTERN
-           re_qt = re.compile(pat % '(?P<name>Qt5.*)' + '$')
+         if Utils.unversioned_sys_platform() == 'darwin':
+                 pat = "%s\.framework"
+         re_qt = re.compile(pat%'Qt5?(?P<name>.*)'+'$')
                for x in dirlst:
                        m = re_qt.match(x)
                        if m:
-                           self.qt5_vars.append(m.group('name'))
+                         self.qt5_vars.append("Qt5%s" % m.group('name'))
                if not self.qt5_vars:
                        self.fatal('cannot find any Qt5 library (%r)' % self.env.QTLIBS)
```

ita1024:
The changes has been applied, so we can close this now.

cawka:
[...]

PR #1912, opened 2017-02-15
qt5: Workaround for QT5 detection on macOS (Bug #1910)

cawka:


ita1024:
Could we leave the regexp as `re_qt = re.compile(pat % '(?P<name>Qt5.*)' + '$')` ? What is the reason to remove Qt5 from the name and add it again?

ita1024:
@fedepell ^^

cawka:
I described the issue in the bug #1912.  There is no "5" in the framework name, unfortunately.

ita1024:
Yes, thank you.

fedepell:
Sorry for the late response. 

Looks awkward that they adapt a quite random library naming scheme :/ But if it is so I guess it is correct then to put the different regex (either just for macOS or in general)

ita1024:
@fedepell :+1:

PR #1914, opened 2017-02-16
Make clang_compilation_database add-on always output complete listings.

DragoonX6:
This change makes sure that if the build would fail and later be resumed that the compile_commands.json would always contain the complete listing of files.
Before this change you would have to do a complete rebuild in such a situation to make sure the file is complete, which is very undesirable on large projects.

One difference is that when you use unity builds, both the unity build file and the files processed by the unity module are present in compile_commands.json. I view this as a feature, as IDEs will have an easier time picking up flags for the appropriate file.


ita1024:
Such a change will force the compilation of all c++ files even when everything is up-to-date. This is likely to be very slow.

I also wonder what will happen in case of compilation errors.

The method `exec_command` is always defined so `getattr(self, 'exec_command', None)` looks redundant.

I will have a closer look at this.

DragoonX6:
>Such a change will force the compilation of all c++ files even when everything is up-to-date. This is likely to be very slow.

Don't worry, what is happening here is that I temporarily override the exec_command function with a function that does nothing if the task status is `Task.SKIP_ME`, after that the original cached in `old_exec_command` is restored.

>The method exec_command is always defined so getattr(self, 'exec_command', None) looks redundant.

So `t.exec_command` should be fine?

>I also wonder what will happen in case of compilation errors.

It looks to be behaving normally, although I'm not too sure on how to properly test this, all I can say for this is "works fine for me".

ita1024:
> after that the original cached in old_exec_command is restored.

I see this. The point is to collect all the "last_cmd" values.

> So t.exec_command should be fine?

I believe so.

notsimon:
I think their might be an issue when no C compiler is loaded.

```
  File ".tools/clang_compilation_database.py", line 73, in <module>
    t = Task.classes[x]
KeyError: 'c'
```

Here's the fix:
```
for x in ('c', 'cxx'):
	if x not in Task.classes:
		continue
```

Should I make another pull-request ?

DragoonX6:
You're right, that was an oversight of mine.

PR #1920, opened 2017-03-07
Extend ListContext (waf list) to print TaskGen descriptions

justinfx:
For more complex projects which contain a larger number of targets, it can be useful to assign descriptions to tasks so that a user can have more understanding of the purpose of each target at a glance.

This merge extends the `ListContext` to also look for a "description" field on each task generator when printing it during the `waf list` command. Descriptions are printing in normal color, aligned to the longest target name.


ita1024:
`getattr(tgen, 'description', None) or ''` is probably better written as `getattr(tgen, 'description', '')`, and 
`self.description` is optional for this reason.

justinfx:
I actually specifically wrote it as `getattr(tgen, 'description', None) or ''` to handle cases where someone passes values like None or empty list or False and have it set to an empty string instead. I felt it would make sure not to format a repr of a non string. 

Also, isn't adding the default description field useful for documentation purposes? 

But thanks for the merge!

ita1024:
1. I believe that such situation is unlikely.
2. Objects add up. Since the description field is not mandatory I would rather save the memory for now.

PR #1925, opened 2017-03-14
Add the ability to detect Visual Studio 2017 to msvc

kulibali:
This PR adds the ability for the `msvc` tool to detect installations of Visual Studio 2017.

Issue #1924, opened 2017-03-14
Support the new MSVC Build Tools


bbalp:
Microsoft now provides a standalone version for C++ (without IDE) called Visual C++ Build Tools. They changed the location where they used to install all the stuff. For example, now they use `C:\Program Files (x86)\Microsoft Visual Studio\2017\BuildTools\VC\Auxiliary\Build` to install the scripts called `vcvars32.bat`, `vcvars64.bat`, `vcvarsall.bat`, `vcvarsamd64_x86.bat`, `vcvarsx86_amd64.bat`.

Also, the Visual Studio 2017’s latest toolset version is vc141. Which, I guess, should become `msvc 14.1` ?

May the Force be with Waf :)

ita1024:
I hope that #1925 fixes this.

kulibali:
#1925 detects installations of Visual Studio 2017 (version 15.0).  I haven't tried it with the Visual Studio Build Tools, either 14 or 15.

bbalp:
I tried on two machines, one with Community 2017 installed, the other with BuildTools 2017. I confirm that it is working with the new VS 2017 (v15.0), but not with the BuildTools variant which installs no registry key to find it. I investigated the problem. Microsoft wrote a fresh article about [Finding installed Visual C++ tools for Visual Studio 2017](https://blogs.msdn.microsoft.com/vcblog/2017/03/06/finding-the-visual-c-compiler-tools-in-visual-studio-2017/):

> (...) we have removed the VS150COMNTOOLS registry key to support the new world where multiple VS instances of the same product can be installed side-by-side on the same machine, we know that many build systems and tools have relied on this in the past and the new options are not an identical replacement (...)

They suggest three ways to find installed instances from now on:
- [vswhere.exe](https://github.com/Microsoft/vswhere)
- [PowerShell API](https://blogs.msdn.microsoft.com/heaths/2017/01/25/visual-studio-setup-powershell-module-available/)
- [VS Setup COM API](https://github.com/microsoft/vs-setup-samples) with a [specific example for C++](https://github.com/adamwelch/vs-setup-samples/tree/findVCtools)

So, I guess a quick fix would be to check the default installation path for VS BuildTools, which is `C:\Program Files (x86)\Microsoft Visual Studio\2017\BuildTools`. But writing something to use the new tools they provided is needed.

[...]

PR #1926, opened 2017-03-15
Update #1925 to correctly configure x64 builds on VS2017

kulibali:
This PR is an update to #1925 that fixes x64 builds not being correctly configured.

Note that as noted in #1924 these changes do not support the MS Visual C++ build tools, just the full VS2017 installation.


bugengine:
It still misses ARM and cross compilers
the following code:
	if os.path.isfile(os.path.join(vc_path, 'VC', 'Auxiliary', 'Build', 'vcvarsall.bat')):
		for target,realtarget in all_msvc_platforms[::-1]:
			targets[target] = target_compiler(conf, 'msvc', realtarget, version, target, os.path.join(vc_path, 'VC', 'Auxiliary', 'Build', 'vcvarsall.bat'))
	

detects all compilers, and is similar to surrounding code.

ita1024:
@bugengine can you make a pull request for this?

PR #1930, opened 2017-03-22
Initial implementation

mortenvp:
Hi,
When running waf on a build-system sometimes things fail. When this happens one can go the the specific build machine and retrieve the log files for debugging. 

I was wondering whether adding support for a `--print-failure-log` or similar could be included into waf.

This would allow us to run `./waf configure --print-failure-log` and see directory in the output what went wrong.

Comments / feedback?

### DIFF
--- a/waflib/Context.py
+++ b/waflib/Context.py
@@ -9,6 +9,7 @@ Classes and functions enabling the command system
 import os, re, imp, sys
 from waflib import Utils, Errors, Logs
 import waflib.Node
+import waflib.Options
 
 # the following 3 constants are updated on each new release (do not touch)
 HEXVERSION=0x1090800
@@ -478,10 +479,18 @@ class Context(ctx):
 		"""
 		if self.logger:
 			self.logger.info('from %s: %s' % (self.path.abspath(), msg))
+		
 		try:
-			msg = '%s\n(complete log in %s)' % (msg, self.logger.handlers[0].baseFilename)
+			logfile = self.logger.handlers[0].baseFilename
 		except AttributeError:
-			pass
+			logfile = None
+			
+		if logfile and waflib.Options.options.print_failure_log:
+			with open(logfile, 'r') as f:
+				msg = '%s\nLog from (%s):\n%s\n' % (msg, logfile, f.read())
+		elif logfile:
+			msg = '%s\n(complete log in %s)' % (msg, logfile)
+			
 		raise self.errors.ConfigurationError(msg, ex=ex)
 
 	def to_log(self, msg):
--- a/waflib/Context.py
+++ b/waflib/Context.py
@@ -9,7 +9,6 @@ Classes and functions enabling the command system
 import os, re, imp, sys
 from waflib import Utils, Errors, Logs
 import waflib.Node
-import waflib.Options
 
 # the following 3 constants are updated on each new release (do not touch)
 HEXVERSION=0x1090800
@@ -485,9 +484,9 @@ class Context(ctx):
 		except AttributeError:
 			logfile = None
 			
-		if logfile and waflib.Options.options.print_failure_log:
+		if logfile and Logs.verbose:
 			with open(logfile, 'r') as f:
-				msg = '%s\nLog from (%s):\n%s\n' % (msg, logfile, f.read())
+				msg = 'Log from (%s):\n%s\n' % (logfile, f.read())
 		elif logfile:
 			msg = '%s\n(complete log in %s)' % (msg, logfile)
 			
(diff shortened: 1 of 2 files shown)


ita1024:
How about https://github.com/waf-project/waf/blob/master/waflib/extras/build_logs.py ?

mortenvp:
Hi @ita1024 
If I understand it correctly it basically logs everything to a single log file?

I think that is different - because we are not really interested in logging everything to file. We are just interested in printing the log file to stdout if something fails.. Because if we do that it will be visible directly in our buildsystem page for that build. 

Does that make sense?

ita1024:
1. While this is only meant for the configuration, this seems to affect all commands
2. The list of command-line options is already quite large
3. Duplicate information will be output (the normal output and then the repeated config.log)

While adding code is easy, making later adjustments is harder so it will take some time to understand the impacts and it is likely that this will be discarded. Subclass configure in your scripts or use your own waf extension if you are in a hurry.

mortenvp:
Hi @ita1024,
A few comments: 
1. I think it would make sense to do for all contexts that have a file backed log. Since if anything fails, we would like to see the output on the build system, without having to retrieve the log files.
2. Yes, might be an issue. 
3. That could be fixed relatively easily.

I think the main show-stopper would be item 2. What do you think is this option too corner case to be part of main waf?

xbreak:
Sorry for butting in and maybe I don't understand the problem, but isn't this easily solved outside waf? Why not simply `cat` the `config.log` (or your custom log file using `extras/build_logs.py`) only on failure: 
`waf configure || cat build/config.log`?

mortenvp:
Hi @xbreak,
Yes, I think there are multiple ways to solve this. That would also be a possibility. 

I think we can say that this is not strictly needed inside waf core so we can close this as a "wont-fix" :)

jhasse:
I really like this!
Instead of adding command line argument I would argue that this should just be the default (one can always `&> /dev/null` if the output is too verbose).

> Why not simply cat the config.log (or your custom log file using extras/build_logs.py) only on failure:
> waf configure || cat build/config.log?

Because this would print out the complete log file, not just the failing part.

xbreak:
@jhasse: That's what the PR proposes as well. And if the full log is not printed you'd miss the context provided by stdout.

mortenvp:
What happens in the PR would not interfer with what is otherwise printed to stdout. It simply "attaches" the current content of the log file to exception raised.On Mar 23, 2017 18:08, xbreak <notifications@github.com> wrote:@jhasse: That's what the PR proposes as well. And if the full log is not printed you'd miss the context provided by stdout.

xbreak:
@mortenvp: Yes, I understood that. I wasn't clear that I was referring to the fact that it would print out the whole log:

>> Why not simply cat the config.log (or your custom log file using extras/build_logs.py) only on failure: waf configure || cat build/config.log?

> Because this would print out the complete log file, not just the failing part.

mortenvp:
One way to avoid adding the additional option, would be to re-use the verbose flag.

We could require `-v` or higher verbosity level.

```
if Logs.verbose:
    # Do the thing
```

What do you think?

jhasse:
Although it might require some more work, I would love if this would work as I misunderstood it first: Print out only the part of the failing test, not the whole log (as this can be achieved with `|| cat build/config.log` as @xbreak pointed out). Then I think it should be the default behavior.

mortenvp:
Yes, would be nice - although in practice I think this would be very hard.. It would require knowing which writes to the log file are related to the thing that eventually failed - not likely to be easy to or very robust. In my opinion.

mortenvp:
I think it makes sense to rely on the verbose flag. I went ahead and changed it to use that instead. 

@ita1024 What do you think? I think that would address your concern 2 and 3.

xbreak:
@jhasse, I agree that it would be nice to get a context sensitive log output. This is e.g. what [Nose](https://nose.readthedocs.io/en/latest/plugins/logcapture.html) does when a unit test fails (and is really valuable). Implementing it in waf on a task-level might be feasible, but is that context enough to warrant the effort spent? What is missing compared to what is printed when a task fails now?

mortenvp:
Just a small ping - what do you think about relying on the verbosity level flag?

jhasse:
I like it. But how does it fix ita's 3. concern?

mortenvp:
That was addressed by changing:

```
msg = '%s\nLog from (%s):\n%s\n' % (msg, logfile, f.read())
```

To:

```
msg = 'Log from (%s):\n%s\n' % (logfile, f.read())
```

I think :)

ita1024:
The output messes with the console colors. While this may not be relevant for server-side builds it is also rather annoying to have this on the -v option (which has other problems, but this is not the topic here). Would something like this work for you?

```diff
                        self.logger.info('from %s: %s' % (self.path.abspath(), msg))
                try:
-                       msg = '%s\n(complete log in %s)' % (msg, self.logger.handlers[0].baseFilename)
+                       logfile = self.logger.handlers[0].baseFilename
                except AttributeError:
                        pass
+               else:
+                       if os.environ.get('WAF_PRINT_FAILURE_LOG'):
+                               msg = 'Log from (%s):\n%s\n' % (logfile, Utils.readf(logfile))
+                       else:
+                               msg = '%s\n(complete log in %s)' % (msg, logfile)
                raise self.errors.ConfigurationError(msg, ex=ex)
```

mortenvp:
@ita1024 Sure that would also work :+1:

ita1024:
Update your copy.

PR #1932, opened 2017-03-27
Fixed Fortran module naming scheme (unexpected upper instead of lower).

haraldkl:


ita1024:
Interesting PRs seem to come after new releases, coincidence? :-)

haraldkl:
In my case: yes. I am sorry, I was late for the last release. I was just looking into the fc module again, because I am right now looking into integrating xevolver (http://xev.arch.is.tohoku.ac.jp), which requires the rose compiler beneath and produces *.rmod files.

Issue #1939, opened 2017-04-04
Xcode6 and build


bestlem:
I tried the xcode6 example in playground.

waf configure and waf xcode6 (and any combination works- you get an Xcode project .

However waf build fails with a python backtrace ending in
  File "/Users/mark/src/cpp/waf/playground/xcode6/wscript", line 43, in build
    tg = bld.framework(
AttributeError: 'BuildContext' object has no attribute 'framework'

Debugging requires more knowledge than I have but the code seems to match the Waf book
But I note when the build runs the bld object is a waflib.Build.BuildContext object and doesn't this need to be an xcode6 object to understand the framework method.

ita1024:
That is a user-contributed which is unsupported. If you have an idea on how to fix the example feel free to submit a pull request.

bestlem:
That is the problem I don't understand the internals of waf well enough.

My first question here is how do you create a BuildContext that is a subclass and not that actiual class.
Is there a supported example of using a subclass of BuildContext

ita1024:
There are examples in the documentation, but start with the following:

1. The xcode module overrides/provides program/stlib which are valid targets (they work for the build) because they are provided elsewhere. The "framework" and other aliases are not provided anywhere and thus cannot be expected to work in builds. The alias "exe" provides one warning for example
2. The example does not detect any c/c++ compiler so the regular targets cannot build anything.
3. The attribute "source_files" is not processed by regular targets ("source" is)

PR #1943, opened 2017-04-09
Fix breaking changes from #1939 where target 'MyExe' was unable to build in XCode

mimon:
In playground/xcode6/wscript: Target 'MyExe' was declared before dependency target 'MyDynLib'.

This caused the generated XCode project to not be able to build 'MyExe' since 'MyDynLib' could not be added as a dependency in the XCode project.


ita1024:
This is strange, the non-xcode build correctly sets the dependency on the lib from the exe. Is there something assuming an order of target declaration in the xcode tool?

mimon:
Yes. It loops all TaskGenerators and then for each taskgen it does the following dependency check:

```python
# Check if any framework/lib to link against is some other target we've made
libs = getattr(tg, 'tmp_use_seen', [])
for lib in libs:
	use_target = p.get_target(lib)
        if should_add_to_xcode:
            ...
```

I guess the better solution is to put that dependency check after all taskgens have been processed, and then add any dependency chain to the project file. But that requires some refactoring which I don't have time for atm.

ita1024:
Can you try the changes above? The target declaration order should not have any impact.

mimon:
Yes, it (d300a7c) builds correctly! 👍

ita1024:
Closing this then.

PR #1945, opened 2017-04-11
Detect MSVS 15.0 Installation path (via vswhere)

refack:
Fixes: #1924

### DIFF
--- a/waflib/Tools/msvc.py
+++ b/waflib/Tools/msvc.py
@@ -431,6 +431,32 @@ def gather_winphone_targets(conf, versions, version, vc_path, vsvars):
 		versions['winphone ' + version] = targets
 
 @conf
+def gather_vswhere_versions(conf, versions):
+	prg_path = os.environ.get('ProgramFiles(x86)', os.environ.get('ProgramFiles', ''))
+	vswhere = os.path.join(prg_path, 'Microsoft Visual Studio', 'Installer', 'vswhere.exe')
+	if not os.path.exists(vswhere):
+		return None
+	args = [vswhere,
+					'-latest',
+					'-products', '*',
+					'-requires', 'Microsoft.VisualStudio.Component.VC.Tools.x86.x64']
+	try:
+		inst_path_lines = conf.cmd_and_log(args).splitlines()
+		ver = None
+		vc_path = None
+		for line in inst_path_lines:
+			parts = line.split(': ', 1)
+			if len(parts) == 2:
+				if parts[0] == 'installationPath': vc_path = os.path.abspath(parts[1])
+				if parts[0] == 'installationVersion': ver = parts[1]
+		if ver and vc_path and os.path.exists(vc_path):
+			version = '.'.join(ver.split('.')[:2])
+			conf.gather_msvc_targets(versions, version, vc_path)
+	except Exception as e:
+		Logs.debug('msvc: gather_vswhere_versions: failure %s', str(e))
+
+
+@conf
 def gather_msvc_versions(conf, versions):
 	vc_paths = []
 	for (v,version,reg) in gather_msvc_detected_versions():
@@ -611,6 +637,7 @@ def get_msvc_versions(self):
 	self.gather_intel_composer_versions(dct)
 	self.gather_wsdk_versions(dct)
 	self.gather_msvc_versions(dct)
+	self.gather_vswhere_versions(dct)
 	Logs.debug('msvc: detected versions %r', list(dct.keys()))
 	return dct
 


ita1024:
This will need some thinking as this raises maintainability questions.

refack:
@ita1024 I think it turned out quite elegant
Also the embeded code is copied almost as is from https://github.com/node4good/msvs-com-helper, I believe it'll stay easy to vendor in in case of bug fixes.
The `msvs-com-helper` code has been tested by `Boost/build`, `nodejs/node-gyp`, `gyp`, `chromium` , it's being considered by `nodejs/node`, and it got a nod from Microsoft, so it's pretty hardened.

ita1024:
The code license is not the one we use here (https://github.com/waf-project/waf/blob/master/waf-light#L5) and the code requires some thinking. I will need some time to understand the implications.

refack:
Updated license.
Feel free to think this over, or ask any questions.

ita1024:
Given that vswhere.exe will be provided along with visual studio, we should be able to remove the dependency on Powershell. We should close this PR and try to make a pure python solution with vswhere.exe in the future.

refack:
@ita1024 replaced COM with a search and call to `vswhere`
Should be easier to swallow ;)

ita1024:
I have had a brief look and there are a few problems:
1. `gather_msvc_detected_versions` is only called if a registry key exists. Since the build tools do not have a registry key, the code is unlikely to be executed
1. Since Visual Studio IDE and the Build Tools may be installed at the same time, it is possible that duplicate results will be returned. How exactly those results will be returned and sorted is unclear. Will the latest version be returned?
1. There are also quite a few magic constants and hard-coded versions. I see a version "15.0" but does that really exist? Is that number not going to be incremented?
1. There are bare `print` statements. Not only they will not work in Python3 but we do have a logger that handles verbosity levels.
1. The subprocess objects `check_output` and `CalledProcessError` are not present in Python 2.5 so this will need to be rewritten
1. The variable `prg_path` may be None and then an exception will be raised. Obtaining the executable path first seems more adequate.
1. The `vswhere_utility.py` file is not really needed.

refack:
> gather_msvc_detected_versions is only called if a registry key exists. Since the build tools do not have a registry key, the code is unlikely to be executed

I forgot about that... Easy to fix

> Since Visual Studio IDE and the Build Tools may be installed at the same time, it is possible that duplicate results will be returned. How exactly those results will be returned and sorted is unclear. Will the latest version be returned?

It's actually worse, you can have multiple version of the same SKU, and multiple IDE etc. That's why they did away with VC15COMMTOOLS.
In this case the `-latest` switch chooses the latest version that answer the requirements.

> There are also quite a few magic constants and hard-coded versions. I see a version "15.0" but does that really exist? Is that number not going to be incremented?

The one in `msvc.py` should be done away with, the one in `vswhere_utility.py` is to be defensive against untested future versions (`Boost.build` asked for that, I can remove it here making it more future proof)

> There are bare print statements. Not only they will not work in Python3 but we do have a logger that handles verbosity levels.

Obv i'll fix.

> The subprocess objects check_output and CalledProcessError are not present in Python 2.5 so this will need to be rewritten

How far back do you want to support?

> The variable prg_path may be None and then an exception will be raised. Obtaining the executable path first seems more adequate.

Ok.

> The vswhere_utility.py file is not really needed.

Ok.

refack:
P.S. I see `subprocess` in the python 2.5 docs https://docs.python.org/release/2.5/lib/node530.html

ita1024:
Ok, but `check_output` is not the same thing as `check_call` and we already have `cmd_and_log` which runs commands and logs configuration errors to `config.log`.

refack:
> and we already have cmd_and_log

fantastic

ita1024:
A new function `gather_vswhere_versions` called in the same manner as `gather_msvc_versions` and executing the subprocess calls should be sufficient to ensure that the latest visual studio version is returned, and fallback to the current behavior otherwise.

refack:
Tried to do everything you asked for. PTAL

ita1024:
Almost there! Please answer the questions below:
1. Are versions returned by vswhere sorted? FYI, the last entry is meant to be the default compiler detected. In `gather_msvc_detected_versions` results are sorted for this reason.
1. Since `vswhere.exe` is the future of visual studio detection, is `self.gather_vswhere_versions(dct)` not supposed to go **after** in order to detect the latest version(s) and fallback to the classic method?
1. Why is `gather_msvc_versions` still modified?

FYI, one could use a logger instead of `print` at the end of the large `Exception` block, but I can take care of that later.

refack:
> Are versions returned by vswhere sorted? FYI, the last entry is meant to be the default compiler detected. In gather_msvc_detected_versions results are sorted for this reason.

Since we use the `-latest` flag it will only return one, the latest installed instance.

> Since vswhere.exe is the future of visual studio detection, is self.gather_vswhere_versions(dct) not supposed to go after in order to detect the latest version(s) and fallback to the classic method?

I wasn't sure what was your selection logic (after detection).
Switched.

> Why is gather_msvc_versions still modified?

Bad merge, removed.

> FYI, one could use a logger instead of print at the end of the large Exception block, but I can take care of that later.

Done.

ita1024:
We are releasing waf 1.9.10 now, and we do not want to risk shipping any broken functionality, so this PR will be merged once we have a proper testing environment.

There is no version of the build tools with vswhere.exe that I can use at this time (and the update install is broken for me), so for all practical purposes this is not something anyone can use right now.

refack:
Sure. Ping me when you're ready, or I'll ping you when I find a stable Buildtools install.

refack:
@ita1024 this you comment and remove? I have a pure python way that triggers `vswhere`.

refack:
NM your e-mail from a week ago just popped up...

refack:
@ita1024 ping me if there's any further work needed or bugs pop up.

ita1024:
It looks good, thank you for your help!

PR #1947, opened 2017-04-15
Replace param 'source_files' with 'group_files'. Simplify code.

mimon:
The `'source_files'` param to the xcode6 tool was originally separated from the
conventional 'source' param because it was used to control how the source files
would appear in the XCode folder UI. Also, it'd allow to add any file extensions,
and not limited to those extensions supported by the loaded set of waf tools.

PR #1939 fixed so that XCode now would pick up source files specified by the `'source'` param.

The next step would be to make the old 'source_files' param an optional param to customize the UI folder organization.

This PR renames (actually aliases) `'source_files'` param to `'group_files'` (i.e. still
compatible with `'source_files'`) to better descibe its meaning. The semantic would be the
following:
```python
src = bld.path.ant_glob('...')
bld(
 source=src, # These are now the files exclusively compiled by XCode
 group_files= { 'Custom Folder': src }, # Optionally customize the way source files appear i the UI (files here are now not considered for compilation in XCode)
 source_files= '...' # Alias for 'group_files' for backwards compatibility
)```


ita1024:
What is the difference between `group_files` and `source_files`? If there is none, why the rename? Else, why not remove `source_files` entirely? This is an "extra" module with no guarantee of compatibility whatsoever.

ita1024:
And what is the purpose of `_filerefs`? `get_fileref_if_exists`? These changes are making the code even less maintainable, so we will not apply this pull request in its current state.

mimon:
> What is the difference between group_files and source_files?
In this commit, they are the same.

> If there is none, why the rename?
Previously, `source_files` was used to collect source files for compilation in XCode, *and* to customize the UI folder structure. In this commit `source_files` is used only to let the user group files in different UI folders (and add additional resource files besides source files). I want to do the renaming to better reflect the param's meaning.

> This is an "extra" module with no guarantee of compatibility whatsoever.
I'd be happy to drop the aliasing if compatibility is not important in this module.

> And what is the purpose of _filerefs? get_fileref_if_exists?
Consider the following:
src = bld.path.ant_glob('src/*.cpp')
bld(
 source=src, # To be compiled by XCode
 group_files= { 'Custom Source Folder': src }, # I want the above source files to appear in an XCode folder called 'Custom Source Folder'
)```

Internally, a `PBXFileReference` is created for each source file in the `source` param. If I want the same physical files to appear in an XCode folder, XCode wants me to reuse the same PBXFileReference object if it exists. Otherwise, some UI features like 'Reveal in Finder' break.

I'd be happy to document this better in the source code for clarity?

ita1024:
This information should be in the source code, and the new usecases should be reflected in the example. The code should also be made more simple, not more complex.

mimon:
Alright. I'll see if I can simplify it. Thanks for the feedback.

mimon:
I've just pushed some modifications. I was able to remove the `unique_filereference()` function in favor for `get_fileref_if_exists()` (which is now named `find_fileref()`). They sort of provided the same functionallity. I've added some doc texts and updated the playground sample to reflect the new changes.

ita1024:
Thank you!

Issue #1950, opened 2017-04-23
Poor performance when -v is passed


jpalus:
In my Linux distribution of choice the preferred way of building packages is with verbose output so it's always clear which compiler gets invoked or what flags are used. For this reason -v is passed in waf based builds but it appears to cause major performance degradation. For example on my machine build of mpv takes 15s (waf build -j8) while with added -v it takes 85s (waf build -j8 -v) which makes pretty significant difference with -v being 5.5 times slower. Not sure though whether it is expected behavior or bug.

ita1024:
This is somewhat expected: the -v option turns on expensive verifications to check the build scripts for errors and mistakes.

Removing some logging may impact mpv positively, but we cannot do much more than that, and builds do not all have the same performance characteristics.

jpalus:
Just reporting back that changes in version 1.9.11 made a huge difference and now verbose output has negligible impact on performance. Thanks!

PR #1952, opened 2017-04-25
Remove undeclared `ve` variable for performance improvement #1950

akira-baruah:
The preceding commit [b37f37a](https://github.com/waf-project/waf/commit/b37f37afcd5e037d0fc149030a7ce7d8e656ff30) removes the declaration of variable `ve`. However, `ve` remains used [here](https://github.com/waf-project/waf/compare/master...akira-baruah:master#diff-9b353364de82dcd69b104ea10dbf8a89L1040) and causes a NameError when running `waf` in verbose mode.


ita1024:
Yes, the post-merge builds were actually red. Thanks.

PR #1959, opened 2017-05-05
Import library files need to be installed into lib/ on Windows

### LINKED ISSUE #1860: .dll.a files should be installed to .../lib/ instead of .../bin/
When creating a shared library on Windows with MinGW Waf correctly creates a `foo.dll` and a `foo.dll.a` file. Both get installed to `${PREFIX}/bin/` though, where the compiler won't find the `.dll.a` file.

Anything I'm missing? Otherwise I'll start working on a PR for this.

jhasse:
Fix #1860.


ita1024:
Nice, thank you!

PR #1964, opened 2017-05-12
Preserve modification time for is_copy=True

jhasse:
For

```python
bld(features='subst', is_copy=True, source='foo', target='bar')
```

`bar` will have the same last modified time as `foo`. Since the `is_copy` feature is mostly used to reconstruct some config files from the src folder to the build folder, having the correct mtime would be nice (also the file is in fact not modified because it has the same content).

Another method could be to use `shutil.copy2` which should even be available in Python 2.5: https://docs.python.org/release/2.5/lib/module-shutil.html That would also preserve the last access time though.

### DIFF
--- a/waflib/TaskGen.py
+++ b/waflib/TaskGen.py
@@ -718,6 +718,8 @@ class subst_pc(Task.Task):
 		if getattr(self.generator, 'is_copy', None):
 			for i, x in enumerate(self.outputs):
 				x.write(self.inputs[i].read('rb'), 'wb')
+				stat = os.stat(self.inputs[i].abspath()) # Preserve mtime of the copy
+				os.utime(self.outputs[i].abspath(), (stat.st_atime, stat.st_mtime))
 			self.force_permissions()
 			return None
 


ita1024:
This is almost certainly going to break existing projects. Overriding the copy function could be made easier though.

jhasse:
> This is almost certainly going to break existing projects.

Are you sure? The modification time was the time the build was run anyway, I can't imagine why someone would need that, in contrast to the modification time of the original file. Last access time is still correct and this only affects `is_copy=True`.

As a bonus using `shutil.copy2` should also be faster then `write(read)`, shouldn't it?

Would you accept this for Waf 2.0?

ita1024:
I cannot imagine why anyone would want to keep the same timestamp either. If the file has not changed, then the file is not copied again.

Why do you think that shutil.copy2 would be faster? Have you tried to measure the gain?

jhasse:
> I cannot imagine why anyone would want to keep the same timestamp either. If the file has not changed, then the file is not copied again.

We're using it to get some .dll files next to our binary. The DLL files are loaded at runtime and the timestamp is used to check, if they have changed. Therefore we need them to be correct.

> Why do you think that shutil.copy2 would be faster? Have you tried to measure the gain?

No, haven't tried that. I just guess, that shutil.copy/copy2 will result in the operation system receiving "copy a to b" which probably has some optimizations at the filesystem level (e.g. copy-on-write, using optimal buffer size, etc.).

ita1024:
1. Your usecase is certainly valid. It does not mean that all applications rely on timestamps in the same way.
2. While enabling `shutil.copy2` is not going to happen - we encountered problems with read-only files - overriding the copy function is an valid concern
3. `shutil.copy2` performance is an amusing argument because:
   - the implementation seems to be in pure python
   - file copy time in projects is usually insignificant
   - when file copy time is significant it is typically best to rely on sub-processes
   - you have not provided any numbers
3. Yes, we can apply the timestamp suggestion in waf 2.0, but that cannot happen in waf 1.9

jhasse:
Okay, I've created a new PR which targets waf-2.0 instead: #1972

Issue #1965, opened 2017-05-12
UnicodeDecodeError when parsing the vswhere's output


bbalp:
Since the commit a65c7c0 an error has appeared when detecting the compilers available with the vswhere's output. This error seems to be related to an encoding issue. vswhere's output is localized according the default locale, not UTF-8. I am seeing this error because my computer run on a French windows and vswhere's is outputting some accentuated characters.

```
Setting top to                           : C:\waf\demos\c
Setting out to                           : C:\waf\demos\c\build
Checking for 'msvc' (C compiler)         : Traceback (most recent call last):
  File "C:\waf\demos\c\waf-1.9.10-f85ace56fb46551039761743214f9116\waflib\Scripting.py", line 122, in waf_entry_point
    run_commands()
  File "C:\waf\demos\c\waf-1.9.10-f85ace56fb46551039761743214f9116\waflib\Scripting.py", line 183, in run_commands
    ctx=run_command(cmd_name)
  File "C:\waf\demos\c\waf-1.9.10-f85ace56fb46551039761743214f9116\waflib\Scripting.py", line 174, in run_command
    ctx.execute()
  File "C:\waf\demos\c\waf-1.9.10-f85ace56fb46551039761743214f9116\waflib\Configure.py", line 85, in execute
    super(ConfigurationContext,self).execute()
  File "C:\waf\demos\c\waf-1.9.10-f85ace56fb46551039761743214f9116\waflib\Context.py", line 88, in execute
    self.recurse([os.path.dirname(g_module.root_path)])
  File "C:\waf\demos\c\waf-1.9.10-f85ace56fb46551039761743214f9116\waflib\Context.py", line 129, in recurse
    user_function(self)
  File "C:\waf\demos\c\wscript", line 21, in configure
    conf.load('compiler_c gnu_dirs')
  File "C:\waf\demos\c\waf-1.9.10-f85ace56fb46551039761743214f9116\waflib\Configure.py", line 155, in load
    func(self)
  File "C:\waf\demos\c\waf-1.9.10-f85ace56fb46551039761743214f9116\waflib\Tools\compiler_c.py", line 23, in configure
    conf.load(compiler)
  File "C:\waf\demos\c\waf-1.9.10-f85ace56fb46551039761743214f9116\waflib\Configure.py", line 155, in load
    func(self)
  File "C:\waf\demos\c\waf-1.9.10-f85ace56fb46551039761743214f9116\waflib\Tools\msvc.py", line 541, in configure
[...]

PR #1966, opened 2017-05-12
Fix decoding of vswhere's output #1965

bbalp:
This fix is working on my French Windows to solve the issue #1965. Can people review and test it, in particular with different locale settings? :)


ita1024:
It does not work in Python3; try the latest from the master branch?

PR #1967, opened 2017-05-16
eclipse: add support also for java project configuration generation

fedepell:


ita1024:
Please add examples in the future. How can we know that this works at all?

fedepell:
Sure no problem, sorry for that! I'll create a playground directory and copy some other playgrounds to make it multi-language and submit a PR.

Just two questions on the extra:
1) Would you mind if I clean up a bit the module? There is some commented code, some unused stuff.

2) Currently (also before my commit was like that) the extra will create all languages profiles for Eclipse. This is good of course for multi-language projects but probably ugly for single language ones. My idea was to create configuration just for the ones used (of course one can rerun it later) to make it cleaner (if you have many you will anyway get them all in that case). But it is a change of past behaviour. What do you think?

ita1024:
1. Please do
2. Ok for me

PR #1968, opened 2017-05-17
eclipse: added a playground area for testing the eclipse extra

fedepell:


ita1024:
This looks like a massive copy-paste.

fedepell:
I know :( The problem is that to demonstrate a bit the function you need at least 2 or 3 modules per language (so you can see that linking between files / classes / objects work) and they should be in separate directories. 

I agree it is anything but pretty. Otherwise I can reduce it to a very small example, but then you would not see that the file dependencies are actually working.

fedepell:
I tried to make it a bit less ugly by at least collapsing the local wscripts for python and c in a single one, so we have a bit less clutter but same demonstration of correct addition of paths to source files.

PR #1984, opened 2017-06-21
eclipse: query the compiler for standard include dirs and add them in eclipse search path, so standard includes can be found also with non standard compilers

fedepell:
if command doesn't succeed use hardcoded paths as before. tried with various gcc and llvm.

### DIFF
--- a/waflib/extras/eclipse.py
+++ b/waflib/extras/eclipse.py
@@ -14,7 +14,7 @@ $ waf configure eclipse
 """
 
 import sys, os
-from waflib import Utils, Logs, Context, Build, TaskGen, Scripting
+from waflib import Utils, Logs, Context, Build, TaskGen, Scripting, Errors
 from xml.dom.minidom import Document
 
 STANDARD_INCLUDES = [ '/usr/local/include', '/usr/include' ]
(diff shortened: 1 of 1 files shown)


fedepell:
This also makes C++ std includes work, as before with only /usr/include and /usr/local/include it would usually not (since /usr/include/c++/x.y.z is missing)

ita1024:
Could you please make the following minor adjustments before I rebase and merge?
1. use `.splitlines()` instead of `split("\n")`
2. catch `Errors.WafError`, not `Exception`
2. use a tiny try-except-else around the call to cmd_and_log, not around 8 lines of unrelated code

fedepell:
Done, hope it's fine now!

ita1024:
There are more things unfortunately:
1. `STANDARD_INCLUDES[:] = []` seems to be the same thing as `STANDARD_INCLUDES.clear()` except that it seems to re-declare a local variable of the same name. Was this your intention or did you mean to modify the value of the constant?
2. The code will fail if the cc/cxx commands ever contain spaces

I have pushed a change to address those matters, please verify if the code still works

fedepell:
Thanks for the cleanups! I did a few runs on my systems (mixed gccs and llvm) and all is fine with the modifications. Thanks and sorry for the messiness :)

PR #1994, opened 2017-07-16
Simplify logic in validate_cfg

lilrc:
I found the code hard to follow when `okmsg` was set unnecessarily almost everywhere so I decided to simplify the logic a bit. While at it I simplified further and made some improvements, IMO.

1. `okmsg` is now set in only two places. If I would preserve old behaviour completely, it would have been only one, but the extra setting place is in `exec_cfg` after a successful modversion. After such the `okmsg` is set to the detected version, which is natural.

2. `msg` is only validated in one block. I think it is now easier to see that `msg` is now always set, regardless of action.

3. The number of exit points of `validate_cfg` is now one instead of three. It should now be more obvious that `validate_cfg` actually has a `switch`-like behaviour. (Here I mean the C keyword, of course.)

4. `validate_cfg` now requires exactly one action (`atleast_pkgconfig_version`, `modversion`, `package`) to be requested. This is reasonable, since anyway only one action will be performed, so better be strict about it.


lilrc:
If you find this PR acceptable, please let me know if you in addition want me to rebase it to waf-2.0.

ita1024:
We cannot apply this on master (waf 1.9), but simplifications are welcome on the waf-2.0 branch.

PR #1999, opened 2017-07-21
Better default uselib_store and define_name in check_cfg.

lilrc:
This fixes the following issues:

Firstly, with the following minimal `wscript`
```
def options(opt):
    opt.load("compiler_c")

def configure(conf):
    conf.load("compiler_c")
    conf.check_cfg(package='alsa >= 1.0.0')
    conf.write_config_header()
```
one ends up with a `config.h` containing this:
```
/* WARNING! All changes made to this file will be lost! */

#ifndef W_CONFIG_H_WAF
#define W_CONFIG_H_WAF

#define HAVE_ALSA_1_0_0 1

#endif /* W_CONFIG_H_WAF */
```
Having the version appended to the define name is really a strange default, since most programs that use third party libraries want something like `HAVE_ALSA` so that the `#if`s do not have to be changed when the dependency is bumped. A more sensible default is to simply take the first part of the package, so that the version specifier is omitted. This applies to `uselib_store` as well.

Secondly, modifying the above script to:
```
def options(opt):
    opt.load("compiler_c")

def configure(conf):
    conf.load("compiler_c")
    conf.check_cfg('alsa >= 1.0.0')
    conf.write_config_header()
```
results in the configuration failing with:
```

Setting top to                           : /home/kalle/test 
Setting out to                           : /home/kalle/test/build 
Checking for 'gcc' (C compiler)          : /usr/x86_64-pc-linux-gnu/gcc-bin/6.3.0/gcc 
Checking for program 'pkg-config'        : /usr/bin/pkg-config 
Checking for 'alsa'                      : not found 
The configuration failed
(complete log in /home/kalle/test/build/config.log)
```
The culprit is that waf reorders the arguments to the below, which is an excerpt from `config.log`
```
Checking for 'alsa'
['/usr/bin/pkg-config', '>=', '1.0.0', 'alsa']
not found
from /home/kalle/test: The configuration failed
```
The behaviour that the two scripts give two completely different results is IMO very counterintuitive. Therefore my suggestion is to completely remove the *k argument extraction magic to improve consistency. BTW, the *k magic is not documented anywhere. This is also more consistent with how `check` works, since it does not do any positional argument extraction.

Previously, this was not a problem since `atleast_version` can be used instead of the above.


ita1024:
I am not sure `Utils.to_list(kw['package'])[0]` is even necessary, but as I see this could be used to replace the `args` parameter?

PR #2003, opened 2017-07-30
GCC: force space between -o and target file.

Ionic:
GCC passes unknown options directly to the linker. "-o$TARGETFILE" would be an example of such an option. If the linker does not support splitting up -o and the target file path by itself, linking will fail.
This happens on OS X, for instance.

Clang is not affected by this, since it does more work in recognizing options and splits up -o and $TARGETFILE before passing it to the linker.

### DIFF
--- a/waflib/Tools/c_config.py
+++ b/waflib/Tools/c_config.py
@@ -1426,7 +1426,7 @@ def multicheck(self, *k, **kw):
 
 @conf
 def check_gcc_o_space(self, mode='c'):
-	if int(self.env.CC_VERSION[0]) > 4:
+	if (int(self.env.CC_VERSION[0]) > 4) or (int(self.env.CC_VERSION[0]) == 4 and int(self.env.CC_VERSION[1]) > 4):
 		# this is for old compilers
 		return
 	self.env.stash()


ita1024:
This fails on quite a few linker configurations. Why do you want that? And cannot you add such thing to your project settings?

Ionic:
What configurations are you talking about specifically? I'd assume `-o $TARGETFILE` is more widely understood by linkers than `-o$TARGETFILE`.

Adding a workaround to a project might work, but is mostly useless, since the bug I'm talking about manifests globally on OS X.

Check this:

```
ionic@nopileos~# cat /tmp/foo.c                             
#include <stdlib.h>

int main (int argc, char **argv) {
  exit (0);
}
ionic@nopileos~# gcc-apple-4.2 /tmp/foo.c -c -o/tmp/foo.o
ionic@nopileos~# gcc-apple-4.2 -v /tmp/foo.o -o/tmp/foo     
Using built-in specs.
Target: i686-apple-darwin13
Configured with: /opt/local/var/macports/build/_opt_local_var_macports_sources_macports.rsync.ionic.de_release_ports_lang_apple-gcc42/apple-gcc42/work/objroot/src/configure --disable-checking --prefix=/opt/local --mandir=/opt/local/share/man --enable-languages=c,c++,objc,obj-c++ --libexecdir=/opt/local/libexec/apple-gcc42 --libdir=/opt/local/lib/apple-gcc42 --includedir=/opt/local/include/apple-gcc42 --program-suffix=-apple-4.2 --with-system-zlib --disable-nls --with-slibdir=/usr/lib --build=x86_64-apple-darwin13 --enable-werror-always --program-prefix=i686-apple-darwin13- --host=x86_64-apple-darwin13 --target=i686-apple-darwin13 --with-gxx-include-dir=/usr/include/c++/4.2.1
Thread model: posix
gcc version 4.2.1 (Apple Inc. build 5666) (dot 3) (MacPorts apple-gcc42 5666.3_15+universal)
 /opt/local/libexec/apple-gcc42/gcc/i686-apple-darwin13/4.2.1/collect2 -dynamic -arch x86_64 -macosx_version_min 10.9.4 -weak_reference_mismatches non-weak -o/tmp/foo -o a.out -lcrt1.10.6.o -L/opt/local/lib/apple-gcc42/gcc/i686-apple-darwin13/4.2.1/x86_64 -L/opt/local/lib/apple-gcc42/gcc/i686-apple-darwin13/4.2.1 -L/opt/local/lib/apple-gcc42/gcc/i686-apple-darwin13/4.2.1/../../.. /tmp/foo.o -lSystem -lgcc -lSystem
ld: unknown option: -o/tmp/foo
collect2: ld returned 1 exit status
ionic@nopileos~# gcc-apple-4.2 -v /tmp/foo.o -o /tmp/foo
Using built-in specs.
Target: i686-apple-darwin13
Configured with: /opt/local/var/macports/build/_opt_local_var_macports_sources_macports.rsync.ionic.de_release_ports_lang_apple-gcc42/apple-gcc42/work/objroot/src/configure --disable-checking --prefix=/opt/local --mandir=/opt/local/share/man --enable-languages=c,c++,objc,obj-c++ --libexecdir=/opt/local/libexec/apple-gcc42 --libdir=/opt/local/lib/apple-gcc42 --includedir=/opt/local/include/apple-gcc42 --program-suffix=-apple-4.2 --with-system-zlib --disable-nls --with-slibdir=/usr/lib --build=x86_64-apple-darwin13 --enable-werror-always --program-prefix=i686-apple-darwin13- --host=x86_64-apple-darwin13 --target=i686-apple-darwin13 --with-gxx-include-dir=/usr/include/c++/4.2.1
Thread model: posix
gcc version 4.2.1 (Apple Inc. build 5666) (dot 3) (MacPorts apple-gcc42 5666.3_15+universal)
 /opt/local/libexec/apple-gcc42/gcc/i686-apple-darwin13/4.2.1/collect2 -dynamic -arch x86_64 -macosx_version_min 10.9.4 -weak_reference_mismatches non-weak -o /tmp/foo -lcrt1.10.6.o -L/opt/local/lib/apple-gcc42/gcc/i686-apple-darwin13/4.2.1/x86_64 -L/opt/local/lib/apple-gcc42/gcc/i686-apple-darwin13/4.2.1 -L/opt/local/lib/apple-gcc42/gcc/i686-apple-darwin13/4.2.1/../../.. /tmp/foo.o -lSystem -lgcc -lSystem
ionic@nopileos~# clang /tmp/foo.c -c -o/tmp/foo.o
ionic@nopileos~# clang -v /tmp/foo.o -o/tmp/foo
Apple LLVM version 6.0 (clang-600.0.57) (based on LLVM 3.5svn)
Target: x86_64-apple-darwin13.4.0
Thread model: posix
 "/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ld" -demangle -dynamic -arch x86_64 -macosx_version_min 10.9.0 -o /tmp/foo /tmp/foo.o -lSystem /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../lib/clang/6.0/lib/darwin/libclang_rt.osx.a
```

OS X's linker doesn't understand `-o$TARGETFILE`, but is able to work just fine with `-o $TARGETFILE`. `clang` actually splits that up in its frontend before passing to the linker - but GCC's frontend is lazier.

Ionic:
To be fair, non-ancient GCC versions handle this situation better by splitting up `-o` and whatever comes after it. The documentation does however explicitly list whitespace between the option and its argument.

I'd still like to be able to use the system's ancient GCC version with waf projects, though.

ita1024:
What prevents you from setting the flags in your projects?

This pull request cannot be applied because:
1. It creates an inconsistency with the other gcc front-ends (g++ & fortran)
2. A configuration test (through `conf.check`) should be added to verify if the -o flag can be separated from the file name

Ionic:
> What prevents you from setting the flags in your projects?

Like I said, nothing in particular, but I would like to rather have a global fix than patching around each software that builds using waf individually.


>This pull request cannot be applied because:
>
>    It creates an inconsistency with the other gcc front-ends (g++ & fortran)

True. I should handle `gxx.py`. Probably the same for `fc_config.py`?


>    A configuration test (through conf.check) should be added to verify if the -o flag can be separated from the file name

At least for GCC, it's rather the other way around - split -o and file name is (or was) the way it expected to get this option. Maybe GCC is special in this regard, you probably have more experience with other compilers than I do.

Checking that sounds sane, though.

To be honest, I have no idea how to do that. Given that Fortran and C/CXX don't have a common base, the code would have to duplicated or put into `Configure.py`, which sounds very wrong, since this is a compiler-specific check. `c_config` also doesn't have any connotation with `fc_config`.

I'm not sure I understand the architecture well enough (or at all.) Would a check need to create a Task and craft a `run_str` to execute? Probably not, since that's how the compiler is invoked later on. I wasn't able to find compilers running checks.

ita1024:
Can you update your copy to 10a2b6bd04ba87f0e7859f61eecee9ba94b70320 ?

Ionic:
4.5 seems to be the earliest version that automatically splits up -o and a file path following it (on OS X anyway.)

Amended commit regarding this optimization.

What about Fortran? Shall we just copy the test to `fc_config`, too? Or just leave it as-is? I've tested `gfortran` and the same thing is happening.

ita1024:
Is it not sufficient to not run the test on gcc >= 5? As for Fortran, I think we can leave this out for the time being. I do not know which compilers (g95? gfortran?) to test and whether those are used at all.

Ionic:
It would be sufficient - but we can certainly break easier, since my tests showed that GCC 4.5 is the first version to support splitting `-o` and the file name in its frontend.

I've also tested `gfortran` and came to the same conclusion (hardly surprising, since it's part of GCC._ Broken for `gfortran` < 4.5 - working for `gfortran` >= 4.5. I haven't tested the `g95` compiler and so far have never installed it. If you're interested, I can do that, though.

ita1024:
Yes, making it work for gfortran would be useful.

ita1024:
Hi, are you going to make a pull request for the Fortran part or should I assume that C/C++ is sufficient and close this pull request?

Ionic:
ACK, will try to do that. Hopefully during the next few days.

Ionic:
Thanks for merging the other PR. Let's close this one up as well, since the original issue is resolved.

PR #2005, opened 2017-08-01
Updates to Xcode6 tool

mimon:
Divided into two commits.

One commit fixes a crash when running waf xcode6 --targets=tg1, tg2.

The other commit makes xcode6 consider the standard build flags used by the other waf c/c++ tools, e.g bld.program(cxxflags=...).


ita1024:
Why is `conf.env.CXXFLAGS='-std=c++11'` necessary? If that is meant for testing, is there any other more useful flag to use?

mimon:
> Why is conf.env.CXXFLAGS='-std=c++11' necessary?
Do you mean the one in playground/xcode/wscript? It is there for testing and demo purposes. Perhaps I should state that in a comment above it?

Is there any other more useful flag to use?
I don't know. Perhaps the '-arch <architecture>' flag is more appropriate?

ita1024:
> Do you mean the one in playground/xcode/wscript? It is there for testing and demo purposes. Perhaps I should state that in a comment above it?

No, i mean that another flag should be used because there is a configuration test for the same flag just two lines below. Having both with the same flags brings confusion.

> I don't know. Perhaps the '-arch ' flag is more appropriate?

How about -O2 if that is the case?

mimon:
Ah, yes, you are right. I completely missed that it was doing a check() below it. I have updated the commit and changed it to -O2 instead and tested it.

ita1024:
I see. Are we expecting to find the `-std=c++11` flag in the xcode project file too then?

Also `conf.env.CXXFLAGS='-O2'` must be written as `conf.env.append_value('CXXFLAGS', ['-O2'])` because the type is supposed to be a list and because existing flags are overwritten.

ita1024:
It would be nice to know whether the `-std=c++11` flags is expected in the xcode project file as well, or just the -O2 one.

mimon:
> I see. Are we expecting to find the -std=c++11 flag in the xcode project file too then? This is important.

Ah, no. You are talking about the helper-mechanics explained in https://waf.io/book/#_configuration_helpers right? That is not implemented yet. Is there any existing functionality in core I can use to retrieve those flags easily?

ita1024:
The method `process_uselib_vars` processes such flags.

mimon:
What method would you prefer to pull them out of env?

For example, atm. I have
`'OTHER_CPLUSPLUSFLAGS': Utils.to_list(self.env['CXXFLAGS']),`

Should I do a search for keys prefixed with CXXFLAGS_* (and repeat for CFLAGS etc.)

ita1024:
Can you rename the method and check that `'OTHER_CPLUSPLUSFLAGS': Utils.to_list(self.env['CXXFLAGS']),` has everything?

mimon:
Sorry, rename which method?

ita1024:
Rename process_uselib_vars to propagate_uselib_vars in xcode6.py

mimon:
Using the playground wscript and changing xcode6.py to:
```
@TaskGen.feature('c', 'cxx')
@TaskGen.after('propagate_uselib_vars', 'apply_incpaths')
def process_xcode(self):
   ...
	print("Value of self.env.CXXFLAGS: " + str(self.env.CXXFLAGS))
	print("Value of self.env.CXXFLAGS_STD11 " + str(self.env.CXXFLAGS_STD11))
```

It outputs
```
Value of self.env.CXXFLAGS: ['-O2', '-fPIC']
Value of self.env.CXXFLAGS_STD11 ['-std=c++11']```

CXXFLAGS does not seem to contain '-std=c++11' ...

ita1024:
Can you add `use='STD11'` to a particular target?

ita1024:
Or update your copy ^ and check if the xcode project files reflect the target-specific std11 flag.

mimon:
Ah of course, we had forgotten to add it to `use`. Now it works. I'll create a new PR

ita1024:
And it looks like it is there. I forgot i could run that project on non-mac computers.

PR #2012, opened 2017-08-18
Store the path the main process had been invoked from in Context.user_dir

captainmalloc:
Waf does change directory as soon as it start executing the main wscript it has been invoked for, c.f. waflib.Scripting.py:waf_entry_point().

Because of this, it is not possible to support relative paths in module options. Those would be relative to the current working directory in which the main waf process is launched from (from the point of view of the user invoking it). 

By storing this path in Context.py, it is possible to write support relative paths in Options:

```
def options(opts):
    currOptsGroup = opts.add_option_group('Custom Flow Options')
    opts.option_groups['Custom_Flow_Options'] = currOptsGroup
    currOptsGroup.add_option('--input-file', action='store', default=None, help="Path to module input file", dest='input_file')

def configure(cnf):
    specifiedInputFile = cnf.options.input_file
    if not os.path.isabs(specifiedInputFile):
            specifiedInputFile = os.path.normpath(os.path.join(waflib.Context.user_dir, specifiedInputFile))

def build(dlb):
    ...

```

### DIFF
--- a/waflib/Scripting.py
+++ b/waflib/Scripting.py
@@ -38,6 +38,10 @@ def waf_entry_point(current_directory, version, wafdir):
 		ctx.parse_args()
 		sys.exit(0)
 
+	# Store current directory before any chdir
+	Context.waf_dir = wafdir
+	Context.launch_dir = current_directory
+
 	if len(sys.argv) > 1:
 		# os.path.join handles absolute paths in sys.argv[1] accordingly (it discards the previous ones)
 		# if sys.argv[1] is not an absolute path, then it is relative to the current working directory
@@ -49,9 +53,6 @@ def waf_entry_point(current_directory, version, wafdir):
 			# TODO abspath?
 			current_directory = os.path.normpath(os.path.dirname(potential_wscript))
 			sys.argv.pop(1)
-
-	Context.waf_dir = wafdir
-	Context.launch_dir = current_directory
 
 	# if 'configure' is in the commands, do not search any further
 	no_climb = os.environ.get('NOCLIMB')


ita1024:
Is it possible to do this without introducing a new variable? `Context.launch_dir` is meant to represent the current working directory, and was abused to become the path to the wscript file for that feature.

captainmalloc:
I proposed a new variable for backward-compatibility, but I agree that using `Context.launch_dir` is a better solution. I'll update the fix.

captainmalloc:
Here a new commit to remove the new introduced variable `Context.user_dir` and fix initialization of `Context.launch_dir`.

ita1024:
Can you rebase your changes? The commit history would look much better.

As for the warning, it simply means that the current working directory is not within the project, so that all targets must be built.

captainmalloc:
About the message, I do not see any implication on running on the build directory and "forcing" all targets or suggesting a "distclean" to the user. Yet I do not think the code is "forcing" the execution of all targets.

ita1024:
A warning is always displayed when the build is executed from a folder that does not belong to the project because building nested projects can give confusing outputs. The waf process is meant to be called from within a project, as the default behaviour is to build all targets below the current working directory.

As I understand it, you want to force all targets to be built when calling waf with a wscript file as a second argument, but without the associated warning. Could you devise a way to have all targets built by default? Setting `def build(bld): bld.targets='*'` can be one way.

captainmalloc:
Let me analyze the behavior from a final user point of view. 

Normally, when executing a waf flow, every target will be built, unless target list is explicitly specified with the **--targets** command line option.

Now, If I execute a waf flow (without **--targets**) from a waf source directory (a directory that contains a wscript), everything works fine. 
If I execute a waf flow (without **--targets**) from a different directory (a directory that's not related to the executing flow), everything works fine but I get the following warning:

`CWD /home/adriano/flow_run_dir is not under /home/adriano/flow, forcing --targets=* (run distclean?)`

_As a user I do not understand what waf is trying to tell me. What is it **forcing** ? I did not specified **--targets**, the execution of all targets is what I asked for! Why is it suggesting to run distclean?_

As a side note, I confirm that Waf is capable of executing flows even if the user current working directory does not contain any wscript: dependencies are correctly defined, tasks are correctly executed, successive builds do update obsolete files as expected and targets selection from command line do behave correctly.

ita1024:
I suspect that your views are limited to the scope of your usecase. Are you happy with the state of this pull request (silencing the warning in your scripts) or are you going to try to make further changes to silence it for your usecase (without breaking current functionality)?

captainmalloc:
I agree with you, probably I am still not seeing all usecases. I think I am satisfied with this pull request right now. 

As you suggested to, I'll add :

```
def build(dlb):
    if not dlb.targets:
        bld.targets = '*'
    ...
```

To silent the warning and still get targets selection from command line.

PR #2013, opened 2017-08-18
Add distclean compatibility with '--no-lock-in-top' command line option

captainmalloc:
When '--no-lock-in-top' command line option is activated, the default implementation of the 'distclean' command does nothing. 

This simple fix proposes a backward compatible solution to make it work if not lock file is available in top directory.

### DIFF
--- a/waflib/Scripting.py
+++ b/waflib/Scripting.py
@@ -302,13 +302,19 @@ def distclean_dir(dirname):
 
 def distclean(ctx):
 	'''removes the build directory'''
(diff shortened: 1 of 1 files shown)


ita1024:
Unfortunately `ctx.options.out` can be empty, maybe it should not be.

Note: when renaming variables, try to give them better names at least :-)

captainmalloc:
On variable names, I didn't want to be disruptive with existing variable names. Next time I'll do better.

PR #2016, opened 2017-08-23
Always allow pre-unpacked waflib.

lilrc:
At jack2 and the other waf projects I work on I use a verbatim copy of waf-light as waf executable and a pre-unpacked waflib (or better put, it has never been compressed). The main reason is that this makes it easier for downstream maintainers [1], which is very desirable. Also it makes git logs look better.
 
So far the only problem I've experienced with this approach is that out-of-source builds fail with:
```
Error: run waf-light from a folder containing waflib
```
This pull request fixes this, by always looking for a pre-unpacked waflib (regardless of the executable name).

[1] https://wiki.debian.org/UnpackWaf


ita1024:
This is not the same thing, as the folder is for a specific Python version. In such case, why not include waf-light verbatim in your project or set the WAFDIR environment variable?

lilrc:
Well, as a user it is probably best to live with a waf-light file or to need to specify WAFDIR in case of out-of-source builds. Seeing a warning, which is (for my use-case) the desired behaviour, is not optimal.

PR #2018, opened 2017-08-27
Make unit testing shell scripts work.

lilrc:
This patch makes the unit test task generator choose target as the executable if there is no link_task, in effect allowing shell scripts (which do not have link tasks) to be unit tested.


ita1024:
1. A task generator method called `to_nodes` already exists, was it unsuitable or was it missed?
2. Why are you assuming that the `target` attribute should be a shell script? Is there no more than one such script to test?
3. Please extend the example in `demos/unit_test` to show how this is supposed to work. Without any example there is no guarantee that the functionality will still work in the future.
4. How about the interpreter? What if a specific interpreter (bash) must run particular tests instead of the default (ksh/dash/cmd.exe)?

lilrc:
1. It was missed. Thanks for pointing that out!
2. The current implementation supports only one executable per test task generator, so this does not change the current behaviour. In this revision `source` can be a shell script too, but `target` is preferred in case i has been substed or built by other means.
3. The examples have been extended with two trivial python scripts that are tested.
4. I revised the PR so that testing any interpreted language would work by adding a `ut_interpreter` member.

fedepell:
Couldn't the existing ut_exec / ut_cmd be used instead of creating a new ut_interpreter ?

lilrc:
@fedepell Both could be used, but they both have their downsides. `ut_exec` has the downside that you would have to restate the test script. Like this:
```
bld(
    feature  = 'test',
    source   = 'test.py',
    ut_exec = bld.env['PYTHON'] + ['test.py']
)
```
`ut_cmd` has the downside that it overrides the user option `--testcmd`.

fedepell:
Hi,
Yes exactly! I would use the first one, or better use ut_str so you can have also the variables expanded (ie. put ${SRC} to refer to sources instead of rewriting it)

lilrc:
@fedepell Good point. I am indifferent whether to write:
```
bld(
    features       = 'test',
    source         = 'test.py',
    ut_interpreter = bld.env['PYTHON']
)
```
or:
```
bld(
    features = 'test',
    source   = 'test.py',
    ut_str   = '${PYTHON} ${SRC}'
)
```
so maybe skipping the `ut_interpreter` is the way to go. I suggest we wait and see what @ita1024 thinks is best. I will re-work the patch in accordance with the most preferred way.

ita1024:
Thanks for adding examples and taking the comments above into account. This looks much better but a few design questions are remaining:

1. The default *cannot* be to use whichever shell script/python interpreter is found first on the system because that makes scripts unportable and tests system-dependent. Using a configuration variable should be required, and tests should fail in case no interpreter is specified (see below).
1. While one target per task generator is the default for task generators producing executables, requiring one task generator per script raises scalability concerns and is likely to make user wscript files more verbose. The method `add_test_results` is not thread-safe but it could be arranged.
1. The `test` feature change causes non-obvious side effects: with such changes `bld(features='py test', source='test.py', ...)` will be broken. It is possible to separate target and test concerns by using dedicated attributes which would also enable several test scripts in a go.
1. The `--testcmd` option is most likely a false problem: that option should have been a configuration parameter because forgetting to specify the interpreter on the command-line is likely.

Example:

```python
bld(
    features = 'test_scripts py',
    source   = 'xyz.py',
    test_scripts_source = 'test1.py test2.py',
    test_scripts_template  = '${TEST_SHELL} ${SCRIPT}'
)
```

xbreak:
I agree that it would be nice if there were better support for interpreted languages in `waf_unit_test`. And if support is officially added I also think it needs to address how the runtime environment for the unit test is set up  and not just use the environment from the current user (makes test results system-dependent). 

If you look at the `pytest` in extras is solves this for Python by doing more less what `waf_unit_test` does plus set up dependent `PYTHONPATH`s.  Currently only `LD_LIBRARY_PATH` is added by `waf_unit_test`.

lilrc:
@ita1024 Okay, this updated patch separates the executable tests and the interpreted tests into different features.
1. The `test_scripts` feature does not do anything unless `ut_str` is set. How to enforce that people do not hardcode programs in `ut_str` (instead of of using configuration variables) I do not know. Maybe there is a better solution, but this at least points people to not using the default shell.
2. The `test_scripts` feature allows multiple scripts in the new `ut_scripts` variable. I also modified `add_test_results` so that there would be no race conditions, by replacing `self.utest_result` with a list named `self.utest_results`.
3. This concern is fixed by splitting out this feature as `test_scripts`.
4. I agree it is nonsense to run valgrind on `/bin/sh`. To solve this we can choose one of the below. Which would you prefer?
4.1. Only use `--testcmd` for executables, leaving the user with no way to run script tests in a special environment (other than valgrind, maybe).
4.2. Create options `--executable-testcmd` and `--script-testcmd` that apply only to the respective features and drop `--testcmd`.
4.3. Create options `--executable-testcmd` and `--script-testcmd` that apply only to the respective feature and keep `--testcmd` which is applied to all tests.

Other changes from the previous patch is that `ut_interpreter` has been dropped in favor of `ut_str`.

@xbreak The `test_script` feature handles the environment the same way as `test` feature, except it does not deal with `LD_LIBRARY_PATH` and its macosx friend. Do you need more sophistication from this?

xbreak:
@lilrc: Consider any non-trivial unit test; it will in most cases have dependencies to the code under test and/or have dependencies to other _taskgens_ or external _uses_. If the test environment is not correctly set the test will either fail or import the wrong code.  Or if you will, turn it around and consider unit tests without any dependencies, how useful can they be? 

So my point is that if support is added for interpreted languages, it would be nice if it includes better support to calculate & set the environment without resorting to what I had to do in `pytest`.

ita1024:
1. The variable `ut_str` is for executables. Re-using it for scripts looks like a bad idea.
2. There is no need to call the variable `ut_scripts` because the system is not the same as for executables. While not strictly necessary, having both features `test` and `test_script` on the same task generator may be necessary in the future.
3. Good.
4. While `--testcmd` must be kept for compatibility reasons, replacing the interpreter by a command-line switch would make no sense. There is no need to add one.

lilrc:
1. Okay, renamed to `test_scripts_template` as suggested before.
2. Renamed to `test_scripts_source` as suggested.
3. (Formerly 4) I moved the `--testcmd` logic from the task to the taskgen to make it explicit that it does not apply to script tests. Some users might be surprised in the future that their test environments are not respected, but that is a nitpick problem since it would have been a problem anyway. (The task's `ut_cmd` code is not reached if `ut_run` is present, which always is the case with scripts.)

For consistency the other `ut_` variables have been renamed to their corresponding `test_scripts_` variables.

@xbreak  The variable `test_scripts_paths` can be used to plug-in extra paths in the environment. For example one can add necessary paths to `PYTHONPATH`. An example of this is given in `demos/unit_test`.

fedepell:
@lilrc the fact with setting PYTHONPATH as you do in test_scripts_path in my opinion is that you have to do it _by hand_. If your module depends on others (via _use_) it would be nice (see pytest extra) that it's built automatically. If you have a rather big project with many dependencies this is quite cumbersome. That is how waf_unit_test creates the LD_LIBRARY_PATH with all the depencies (use) and not by hand.

I have in my repo a generic way of adding paths to the waf_unit_test that may be interesting to you or @ita1024 which I didn't do a PR yet since I was in doubt, in case if you want to give a look:
https://github.com/fedepell/waf/commit/0d955cb73fb68d42dc20a27323a6bee0f5439d83
I use this to pass (by appending/updating) extra env variables, while otherwise it had to be done by rewriting a complete ut_env from scratch.

Why test_scripts_shells when ut_shell does the same? (and as the code is now they are separate anyway)

The novelty of the modification I see is the iteration over the sources. Couldn't this be directly in the waf_unit_test (via option or something) and not just for scripts at this point?

lilrc:
@fedepell The functionality I am trying to implement is for interpreted languages in general, and shell scripts in particular. If this were to implement automated `PYTHONPATH` appending, then it could equally well implement Ruby's `LOAD_PATH`, Lua's `LUA_PATH`, and the corresponding for each other interpreted language that springs to mind.

`test_scripts_shell` is invented for the exact same reason as `test_scripts_template`. `test_scripts_shell` does the same thing as `ut_shell`, but for scripts, and `test_scripts_template` does the same thing as `ut_str` but for scripts. The `test_scripts_` prefix is there simply for consistency, i.e. it avoids mixing `test_scripts_` with `ut_`.

Iteration over the sources is not possible for executables, because each TaskGen generally only produces one executable. Since the scripts need not be compiled, iteration is possible.

fedepell:
@lilrc: exactly what i was asking is: why don't we make it more generic so one can "plug" in depending ie. on the scripting language some environment?

My comments are just in this direction, try to make it a more generic case. waf_unit_test was first just for C/D. Why not make it a generic infrastructure?

lilrc:
@fedepell We could try to make it generic. For example by having a feature that adds the dependent nodes and populates `test_scripts_paths`. This would require some work though... Maybe it could be done in two steps, by first getting the basic test script code in place and then introducing some more fancy logic.

xbreak:
@lilrc: Yes, exactly. Maybe something like this (on a conceptual level)

Split up waf_unit_test into two steps/methods
1. Find dependencies and set LD_LIBRARY_PATH like it does today
2. Create task(s)

Since in this way it would allow other features to insert themselves between 1 and 2 to e.g. modify the environment add dependencies and whatever necessary without re-implementing what step 1 and 2 provides by default.
 
What's not clear to me now is how best to handle "source as test-target" cases. If the task gen has a `link_task` I guess that can be used by default as test-task input.

## Concept

	@feature('test')
	@after_method('apply_link', 'process_use')
	def process_test(self):
		# Recursively find all uses and store in self.use_seen
		# Also, populate a default ut_env with LD_LIBRARY_PATH,


	@feature('test')
	@after_method('process_test')
	def make_test(self):
		# Create the test tasks using the properly set up ut_env.
		# Depending on wether we have a link_task or not we can deduce if the unit test
		# is compiled or not. This is the major draw-back possibly, and another attribute ##
		# could be used to provide the ${SRC} in the tasks.. 
		# Also split up into seperate tasks for each src if required (ut_one_test_per_source)
		
		# The created tests should only have inputs, so in ut_str only ${SRC} will be used.


	
	@before_method('make_test')
	@feature('pytest')
	def add_pypaths(self):
		# A separate feature used to process self.use_seen and update ut_env with Python specific paths...
		pass


	@before_method('make_test')
	@feature('rbtest')
	def add_rbpaths(self):
		# A separate feature used to process self.use_seen and update ut_env with Ruby specific paths...
		pass

## Example use-cases

	# Self-contained unit test executable
	bld(features = 'test cxx cxxprogram',
		source   = 'foo.cpp')
		# ut_str is not needed for embedded test-runners

	# Test runner that loads test from shared object
	bld(features = 'test cxx cshlib',
		source   = 'foo.cpp')
		ut_str   = '${LIB_TEST_RUNNER} ${SRC}')
		
	# For interpreted uses we need to specify ut_str to contain the interpreter _executable_ or test runner or whatever.
	bld(features   = 'test pytest',
		source_TBD = 'my.py test.py files.py',
		ut_str     = '${PYTHON} ${SRC}')	

	bld(features   = 'test rbtest',
		source_TBD = 'my.rb test.rb files.rb',
		ut_str     = '${RUBY} ${SRC}')
		
	# Trivial shell tests do not need any feature to help populate the environment.
	bld(features   = 'test')
		source_TBD = 'my.sh test.sh files.sh'
		ut_TBD_ATTRIBUTE_THAT_SIGNALS_THAT_ONE_TASK_PER_TEST_FILE_SHOULD_BE_CREATED = True, # To signal we want one task per test file
		ut_str     = '${SHELL} ${SRC}')

lilrc:
I really like the concept, I must say. This would however require that `ut_str` is used for both scripts and executables which @ita1024 was sceptical to. You can get rid of the `_TBD_` variables above by always storing scripts in a variable called `scripts`, which can explicitly iterated over since it does not contain any sources that must be compiled and linked. Implementing this would require re-working this commit again, though, but let's see what @ita1024 says before doing some crazy stuff.

ita1024:
(I will read this over the weekend, sorry for the delay)

ita1024:
Script paths will be necessary (PYTHONPATH, etc) at some point, and it is not really clear how this can be achieved with the current system. My concern is that we may have to rework the API and the attributes a little more.

A few details also need some more care (`handle_ut_cwd` should have `@taskgen_method`, there are `except:` blocks catching keyboard events, ...), but I can address those later on.

I can merge the pull request as it is now, but I will mark the functions as "work in progress" and move the unstable examples to an area under the `playground` folder. Would that work for you?

lilrc:
I put `@taskgen_method` and narrowed the `except:` block, so you don't have to do it.

Yes, merging and marking this as WIP works for me.

PR #2020, opened 2017-08-29
Default ignore Vim swap and Python Compiled files

captainmalloc:
Python compiled  `*.pyc` and Vim `*.swp` are quite common, I think Waf can safely include them in default exclusion list.


ita1024:
There are cases where we want to list `.pyc` files, so this change will not happen. The `.swp` suggestion looks like a good idea.

Issue #2017, opened 2017-08-25
Split Python packages problematic for unit testing


xbreak:
Hi,

This is a question on how to best deal with Python unit tests with build time dependencies that you want to execute in the waf context (no installation). I'd appreciate your input and support on this.

I started to improve the `pytest` feature to add support for library dependencies to be able to populate `LD_LIBRARY_PATH` and `PYTHONPATH` for `pyext` and normal libraries correctly so that unit tests for e.g. a Python package that depend on a C extension would work.

During this work I came across the general problem with unit testing in Python where if you create a Python extension or generate Python code that should be installed inside a Python package we currently end up with a split package where the generated parts are located in _build_ and the rest remain in _source_, only to be merged when installed. 

For example, consider a package `foo` that wraps the extension `foo.foo_ext`, this is what you could end up with (see build def [1]):

    .
    |-- build
    |   `-- src
    |       |-- foo
    |       |   `-- foo_ext.cpython-35m-x86_64-linux-gnu.so	# 'foo_ext' is correctly located inside the 'foo' package by specifying "target='src/foo/foo_ext'"
    |       `-- foo_ext.c.1.o
    |-- src
    |   |-- foo
    |   |   |-- foo.py	# "from foo import foo_ext" -> ImportError: cannot import name 'foo_ext'
    |   |   `-- __init__.py
    |   `-- foo_ext.c
    |-- test
    |   `-- test_foo.py
    `-- wscript

Since the package is split between _build_ and _source_ this does not work until they are merged. 

The project I'm working on will write unit tests in Python for code that has generated dependencies like C-extensions, Protobuf code, Qt resources/mocs to name a few, so this is something I'd really like to solve. 

How can we best support this? Could e.g. the `py` feature (optionally maybe) copy sources to _build_ to make the package complete there? This would be the simplest case for `pytest` to support since it would just be a matter of setting the correct `PYTHONPATH`.

Otherwise `pytest` would have to do it somehow. Although it's not clear to me at the moment how best to do it. Should it e.g. install its `py` dependencies to its own build node or merge "in situ"? And since the information on how to merge a package is not really explicit it might be problematic.

Thanks!

[1]:

    def build(bld):
        # Python C extension
        bld(name         = 'foo_ext',
            features     = 'c cshlib pyext',
            source       = 'src/foo_ext.c',
            target       = 'src/foo/foo_ext')

        # Python package under test
        bld(name         = 'foo',
            features     = 'py',
            use          = 'foo_ext',
[...]

PR #2022, opened 2017-09-04
pytest: Added better support for normal and pyext library dependencies.

xbreak:
This also includes the `buildcopy` feature that allows construction of complete Python packages in the build dir.

See `playground/pytest/wscript` for use-cases and #2017 for motivation why `buildcopy` was needed.


ita1024:
There is already a task generator method called `to_nodes`. Be careful with `shutil.copy2` and opening/closing files in `Python < 3.6`; prefer `Utils.readf/writef` instead.

xbreak:
Thanks!

I found another corner case for `pytest` to handle, so while I'm fixing that I'll also rename `to_nodes` to `to_src_nodes` to avoid any confusion.

Regarding `shutil.copy2` I'm not familiar with the risks you seem to be referring to, could you clarify?

ita1024:
File handles are opened exclusively and inherited. If a file is being written while other tasks are running, then those may prevent the file from being read by other processes.

xbreak:
For reference, this seems to cover the info about the issue:
- [Issue 17634: Win32: shutil.copy leaks file handles to child processes](https://bugs.python.org/issue17634) and
- [PEP 446 -- Make newly created file descriptors non-inheritable](https://www.python.org/dev/peps/pep-0446/)

Since `Utils.readf/writef` puts the whole file in memory I started writing a `Utils.copyfile` function which use bounded memory instead. Now, it would be nice if file attributes were preserved during the copy, but here it seems like there's no reasonable way to do that using file descriptors opened with `os.O_NOINHERIT` (the `shutils.copystat` function seem very messy to implement in a portable way).
On the other hand, for the `buildcopy` use case I don't think the attributes are important unless those files are accidentally picked up by `<task_gen>.to_nodes` somewhere.

At the same time the same general problem with inherited FDs seem to be around in other places as well in waf (the install tasks and file attribute modifications for example) so if it's really a problem we want to solve then it's going to impact other places as well.

Reading tickets about the issue it seems most likely to impact you if a subprocess holds a file handle to something in a directory which some other process tries to delete. Is this likely to happen? Not in the `buildcopy` use case I think.

PR #2026, opened 2017-09-11
When '--top' is specified there is no reason to climb

captainmalloc:
All in title.


ita1024:
I am not sure I understand: if `top` is set then the inside of the loop `while cur and not Context.top_dir` is never reached.

Could you elaborate or perhaps produce a test in:
https://github.com/waf-project/waf/blob/waf-2.0/tests/init/wscript#L121

captainmalloc:
Sorry about that, I forgot to remove that part of the `while` condition. 

My idea is to discover is lockfile exists even if the `--top` option has been defined. With this  previous `Context.out_dir` can be restored, if a lock file exists. Does it make sense?

ita1024:
In this case `start_dir` is also affected. Two tests were added to account for that.

ita1024:
Please update your copy and try it.

captainmalloc:
Everything seems to work fine, thanks for merging this

PR #2027, opened 2017-09-13
Remove rendoundant commad line options

captainmalloc:
Now that command line options parsing is more stable and `--top` behavior has been completely cleared, I think supporting `wscript` file passed as a first argument is completely redundant with `--top`.

I do not know the historical reasons for this options to exists, this pull request is to actually propose a cleanup for `waf-2.0`.

### DIFF
--- a/waflib/Scripting.py
+++ b/waflib/Scripting.py
@@ -38,17 +38,6 @@ def waf_entry_point(current_directory, version, wafdir):
 	start_dir = current_directory
 	no_climb = os.environ.get('NOCLIMB')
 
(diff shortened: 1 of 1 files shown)


ita1024:
No, we cannot do that, it was introduced in #1517 and there was one related request on the mailing-list a few weeks ago.

ita1024:
I will close this, unless you have a practical reason for requesting this.

captainmalloc:
Thanks for pointing that out, there is no practical reason to remove this, it was just related to code cleanup.

ita1024:
Ok!

PR #2028, opened 2017-09-14
protoc: extend extra to support also Python language

fedepell:


ita1024:
This is a good idea but there are a few problems:
1. `ext_out = ['.h'..` is meant for the build order. Search for lines mentioning `ext_in = ['.h'...`. This is not strictly necessary for python compilation because file dependencies are strictly explicit (but you may want to verify this).
2. How about targets having both `c++` and `py` features? Could we do both in one step? If no, why?
3. There is no validation that `protoc_include` contains valid paths
4. The name `protoc_include` does not end with an `s` so this is inconsistent with `includes`

fedepell:
Many thanks for the feedback! I added a commit right now, so:

1) Fixed. I added also for Python since I saw other places doing so even if as you said it should not be needed.
2) I reshuffled a bit the code and implemented it so both can be done at the same time in one single step since the tool supports multiple language output. I kept it backwards compatible.
3) What kind of validation do you suggest? I checked a few others doing similar stuff (qt5, ccroot) and see that variables are usually just used as is. Please just let me know and I can add.
4) Fixed, sorry for this!

I plan to do the same also for Java (just requires a bit of .proto parsing) so I kept the structure open for that already.

Hope it's better now, thanks again!
F.

ita1024:
This is better, but as I understand, `PROTOC_FLAGS` is meant to be a list, and in this case there is no reason to assign strings to it. A few more changes will be made.

fedepell:
Thanks for merges and follow up fixes!

PR #2034, opened 2017-09-19
Node.ant_glob: Do not ignore case by default, as specified by the docs

xbreak:
Found possible regression in the `Node.ant_glob` `ignorecase` parameter. It is documented to be default `False` but acts as default `True`.

Added regression test case in `tests/apis`.

Regression point: 4226d0cf7bbc3e15cbda02cd13adb693fc60751b [diff](https://github.com/waf-project/waf/commit/4226d0cf7bbc3e15cbda02cd13adb693fc60751b#diff-5a81127acf8ec2d098d2b62665a140b4L621)


ita1024:
This is much better. Thank you.

Issue #2035, opened 2017-09-19
Accidental cross-context "includes" list sharing


daniel-brosche:
I have found an issue and I was able to reproduce them by extending the playground/use example.

https://github.com/daniel-brosche/waf/commit/08f560e465158e82a0ada08b9e05897f606cdb7a

When I build this example with: 
```
waf-light configure build install --destdir=/tmp/
```

I get this error:
```
Traceback (most recent call last):
  File "/home/brosche/Dev/waf_fork_github/waflib/Scripting.py", line 157, in waf_entry_point
    run_commands()
  File "/home/brosche/Dev/waf_fork_github/waflib/Scripting.py", line 242, in run_commands
    ctx = run_command(cmd_name)
  File "/home/brosche/Dev/waf_fork_github/waflib/Scripting.py", line 226, in run_command
    ctx.execute()
  File "/home/brosche/Dev/waf_fork_github/waflib/Scripting.py", line 564, in execute
    return execute_method(self)
  File "/home/brosche/Dev/waf_fork_github/waflib/Build.py", line 231, in execute
    self.execute_build()
  File "/home/brosche/Dev/waf_fork_github/waflib/Build.py", line 251, in execute_build
    self.compile()
  File "/home/brosche/Dev/waf_fork_github/waflib/Build.py", line 345, in compile
    self.producer.start()
  File "/home/brosche/Dev/waf_fork_github/waflib/Runner.py", line 398, in start
    self.refill_task_list()
  File "/home/brosche/Dev/waf_fork_github/waflib/Runner.py", line 244, in refill_task_list
    tasks = next(self.biter)
  File "/home/brosche/Dev/waf_fork_github/waflib/Build.py", line 797, in get_build_iterator
    self.post_group()
  File "/home/brosche/Dev/waf_fork_github/waflib/Build.py", line 763, in post_group
    tgpost(tg)
  File "/home/brosche/Dev/waf_fork_github/waflib/Build.py", line 741, in tgpost
    f()
  File "/home/brosche/Dev/waf_fork_github/waflib/TaskGen.py", line 239, in post
    v()
  File "/home/brosche/Dev/waf_fork_github/waflib/Tools/ccroot.py", line 124, in apply_incpaths
    self.env.INCPATHS = [x.path_from(cwd) for x in lst]
  File "/home/brosche/Dev/waf_fork_github/waflib/Node.py", line 468, in path_from
    if c1.parent:
```
However when I use the command seperately like:

```
waf-light configure build
```
[...]

PR #2036, opened 2017-09-20
Extended playground use example

daniel-brosche:
Related to https://github.com/waf-project/waf/issues/2035


ita1024:
Can you briefly describe what that example is supposed to show and why we are supposed to have it? Thanks.

daniel-brosche:
It extends the use example with some more complexity which means it recurses another hierachy down and extends the dependency graph. 
My main idea was to provide this code for reproduction and regression testing.
I have noticed that there is still a comment in the code that could be removed before merging.

Are you interested in more playground examples?

I have also some another scenarios that are maybe interesting (as mentioned more complex multiproject scenarios).

ita1024:
I am not particularly interested into more examples unless they have good quality as this means more maintenance.

daniel-brosche:
I understand, I will create some templates and store them in my repository. Maybe you can reference them later on your website.

PR #2044, opened 2017-09-22
make variants an integral part of waf

rpuntaie:
Variants are so common that one should not need to resort to solutions by coding in the wscript.


ita1024:
We used to have something like that around waf 1.4-1.5, and we did remove it because such assumption does not really hold.

Can you elaborate on why you believe that `os.path.normpath(self.variant)` is necessary?

rpuntaie:
Regarding `normpath`: with a variant a/b, on Windows the output is `..\build\a/b` instead of `..\build\a\b`

Regarding `variants`: if `variants` is not in wscript, this variant implementation is disabled, like nonexistent.  It therefore cannot be an obstacle to more advanced users. They can still go for its own solution.  New users will most likely not do their own implementation and will want an existing solution. It was like that for me.

What I also missed, is a way to copy files to the build folder. I've recently detected `buildcopy` in extras. This should be part of the main waf, too. Maybe use shorter names `bld_cp` and `files` instead of `buildcopy_source`. Don't use `source`, because that has a special meaning.

Regarding the --pdb option: I'm used to this from `py.test`. It helps for advanced users.

ita1024:
Is there a problem using `/` instead of `\`?

For copying files, see `demos/subst`, on the `is_copy` part.

rpuntaie:
Regarding / instead of  \: No problem, it just looks better. That is not what the pull request was about.

ita1024:
Making variants the default is not going to happen because we moved away from them not a long time ago. We know that variants are never done in a single way, and even the examples in `demos/variants` show that there is more than one possible design.

The `os.normpath` or the `pdb` changes look interesting, but they are mixed with everything else. Similar changes could be applied from our side you are not going to make a pull request.

Note: no API will be changed or broken until Waf 2.1.

PR #2049, opened 2017-09-26
protoc: add java support for protobuffers

fedepell:


ita1024:
In the future please extract variables or file names instead of copy-pasting long expressions such as `[node.abspath().rfind(os.sep)+1:node.abspath().rfind('.')]`

PR #2055, opened 2017-10-13
also use DEST_OS in cfg_cross_gnu

rpuntaie:
I've added an illustrative example + considered DEST_OS in the code

### DIFF
--- a/waflib/extras/cfg_cross_gnu.py
+++ b/waflib/extras/cfg_cross_gnu.py
@@ -81,6 +81,13 @@ This example ``wscript`` compiles to Microchip PIC (xc16-gcc,.. must be in PATH)
 				else:#configure for pc SW
 						...
 
+		def build(bld):
+				if 'fw' in bld.variant:#firmware
+						bld.program(source='maintst.c',target='maintst');
+						bld(source='maintst.elf',target='maintst.hex',rule="xc16-bin2hex ${SRC} -a -omf=elf")  
+				else:#build for pc SW
+						...
+
 """
 
 import os
--- a/waflib/extras/cfg_cross_gnu.py
+++ b/waflib/extras/cfg_cross_gnu.py
@@ -16,14 +16,17 @@ The variables are obtained from the environment in 3 ways:
 2. By defining HOST_x
 3. By defining ${CHOST//-/_}_x
 
+else on can set ``cfg.DEST_OS``.
+
 Usage:
 
 - In your build script::
 
     def configure(cfg):
       ...
-      conf.load('c_cross_gnu')
+      conf.load('cfg_cross_gnu')
       for variant in x_variants:
+		setenv(variant)
         conf.xcheck_host()
         conf.xcheck_host_var('POUET')
         ...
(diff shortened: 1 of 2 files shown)


ita1024:
There are lots of changes but I am not sure of the clarification value. And why `envar = os.environ.get(name) or conf.env.DEST_OS`?

rpuntaie:
I've found that cfg_cross_gnu cannot be loaded together with gcc, because the latter will choose the local system as DEST_OS. So I had to do the configuration steps in gcc inside the wscript configure. I would like move these steps into cfg_cross_gnu, too. 

cross_gnu would be a better name. 

There should be a way to tell the cross_gnu tool the destination system inside wscript. I have chosen DEST_OS for this. This way one can compile for different DEST_OSes in one wscript via variants.

Of course, a better proposal is always welcome.

rpuntaie:
I've done the mentioned changes.

ita1024:
* `xcheck_envar` returns `DEST_OS` by default and that is the wrong place to do such a thing
* The docstrings now contain mixed indentation (both tabs and spaces)
* The docstrings contain project-specific flags of little documentation value

I could apply such changes but the above will be obliterated.

rpuntaie:
fixed 1 and 2

regarding 3: fixed partially. I would like to keep the settings for Microchip PIC in the docstring in order for someone to find it, if he wants to compile with waf and
- to make up for not having a separate tool 
- as an example for other MCUs.

ita1024:
The module is assuming gcc, but cross-compilation can also happen with g++, and with the clang family, so the `configure` function is incorrect.

rpuntaie:
Since it is called cross_gnu, it should not include clang. gcc can also compile C++. I didn't add g++, because I had no need for it and it was not there in the first place. Many cross compilation tool chains don't include g++. Nevertheless one could and maybe should add g++ to cross_gnu.

zougloub:
Hi, sorry guys I was attending some personal stuff (and still am), but here are some concerns of mine:
- These changes are breaking the API of the module which may be used in the wild (well, at least by me);
- Some lines of the patch are not changing anything;
- The "cfg" naming was considering that the module contains functions used for configuration (like its friend `cfg_altoptions`) and the intent was that to allow maximum flexibility vs. hard-coding a particular way of doing things;
- The "gnu" naming is coming from the name of the environment variables that are used, not relating to the compiler, and it was specifically supporting clang (from the docstring).

I would advise to split the module so the specialization is done by the `cross_gnu` tool while the free functions are in `cfg_cross_gnu`.

rpuntaie:
Generally one can not have eternally the same interface, because it would block the evolution of a system.

The idea of tools in WAF is to load them in `configure()`. cfg_cross_gnu did not do that.

The interface of the functions have been extended in a backward compatible way.
It is just that now `configure()` is called when you load the tool inside the wscript `configure()`.
If you don't want that you can do an import at the top of wscript.

Through these changes the usefulness of this tool has been extended immensely.

Since the name of the tool ends in `_gnu`, the user expects it to be for gnu tools.
Nevertheless this interface has not changed and you can continue to define 
`HOST_CC="clang -..." waf configure`  environment variables.

rpuntaie:
For later maybe WAF can support cross compilation in a more fundamental way, i.e. via the actual tool files like `gcc.py`, `gxx.py` , `gas.py`, ...

PR #2056, opened 2017-10-14
Add a find parameter to to_nodes.

karllinden:
I found that it is impossible to only search the source directory for nodes when invoking to_nodes, because `to_nodes` always uses `find_resource` which always looks in the build directory first. This pull request makes it possible to substitute `find_resource` against for example `find_node` so that only the source directory is searched.


ita1024:
The method `to_nodes` is a critical spot for performance in many applications. Could you pass Node objects to the method or use your own `to_nodes_x` method instead?

lilrc:
I have no performance figures, but does this make the function considerably slower?

I am writing a waf module for a tool, so I cannot control what the consumer build systems will pass into it, so the only sensible thing is to roll another function that uses `find_node` instead. Do you want to include a more generic function in `waflib` or should I just carry it myself?

ita1024:
The method `to_nodes` is typically called in extensions. Since you are writing a waf module, then that module can surely provide its own function(s) without a need to add/alter `waflib`?

I am also eager to know what the case for searching files in the source directory instead of the build directory. Is this meant for replicating files to the build directory? In such a case, why not leverage the `subst` examples directly?

rpuntaie:
I suggest you give the `example.c` in the source path a different name, e.g. `example.c.tst`. Then you create two target files `example.c` and e.g. `test_example.c`. I comment here because I have a similar approach: I start with a python template file (`xxx.c.tpl`) and create the `c` and the test file from it.

Issue #2061, opened 2017-10-17
Intel C++ Compiler Detection when using Intel Fortran Compiler


Schildkroete23:
I tried to build the Fortran example using the Intel Fortran Compiler on Windows OS.
But the Intel Fortran compiler cannot be found.
In the config.log I get the message: „could not find the program ['ICL']“
It seems like waf is searching for the Intel C++ Compiler (icl.exe)
when trying to detect the Intel Fortran Compiler (fort.exe).
Unfortunately the Intel C++Compiler is not installed on the machine I am using
and there are no plans to get it.
Why is waf searching for the icl.exe for detecting the Intel Fortran Compiler?

ita1024:
No idea but that used to work. Try perhaps the following and open a pull request if it works?

```diff
diff --git a/waflib/Tools/ifort.py b/waflib/Tools/ifort.py
index 20e9433..89a1651 100644
--- a/waflib/Tools/ifort.py
+++ b/waflib/Tools/ifort.py
@@ -314,7 +314,7 @@ def _get_prog_names(self, compiler):
                lib_name = 'XILIB'
        else:
                # assumes CL.exe
-               compiler_name = 'CL'
+               compiler_name = 'FORT'
                linker_name = 'LINK'
                lib_name = 'LIB'
        return compiler_name, linker_name, lib_name
```

Schildkroete23:
This does not work, but it was a good hint were I have to look at. Thanks.
So I could figure out that the following works to detect the Intel Fortran Compiler on my machine.
```diff
diff --git a/waflib/Tools/ifort.py b/waflib/Tools/ifort.py
index 20e94332..bf8ac7b6 100644
--- a/waflib/Tools/ifort.py
+++ b/waflib/Tools/ifort.py
@@ -137,9 +137,9 @@ def gather_ifort_versions(conf, versions):
                        except OSError:
                                pass
                        else:
-                               batch_file=os.path.join(path,'bin','iclvars.bat')
+                               batch_file=os.path.join(path,'bin','ifortvars.bat')
                                if os.path.isfile(batch_file):
-                                       targets[target] = target_compiler(conf, 'intel', arch, version, target, batch_file)
+                                       targets[target] = target_compiler(conf, 'ifort', arch, version, target, batch_file)

                for target,arch in all_ifort_platforms:
                        try:
@@ -148,9 +148,9 @@ def gather_ifort_versions(conf, versions):
                        except OSError:
                                continue
                        else:
[...]

PR #2063, opened 2017-10-18
Use correct batch filed and function for Intel Fortran detection on Windows

Schildkroete23:
Fix for #2061


ita1024:
Excellent, thank you!

PR #2066, opened 2017-10-23
Change gcc_modifier_ function finding logic to use underscores.

DragoonX6:
This change allows you to create `gcc_modifier_*` functions for CHOSTs like `x86_64-w64-mingw32`.  
Which would change the (invalid) searched function from `gcc_modifier_x86_64-w64-mingw32` to `gcc_modifier_x86_64_w64_mingw32`.  
Which is a valid python function signature.


DragoonX6:
[Rejected]

PR #2071, opened 2017-11-07
Set PYTHONPATH for pytest-tests depending on a pyext target

kljohann:
By not only including the appropriate build folder in `LD_LIBRARY_PATH`
but also in `PYTHONPATH`, the python extension module can be imported
by the test script.

For example the following will now work using `waf build` or `waf --alltests` (and failed to work before this change):

#### `wscript`:
```python
def build(bld):
    bld(
        target='some_extension',
        source='...',
        features='cxx cxxshlib pyext',
        use=['...'],
        install_path='${PREFIX}/lib',
    )

    bld(
        pytest_source='some_extension_test.py',
        features='pytest',
        use='some_extension',
        ut_str='${PYTEST} -q ${SRC}',
    )
```

#### `some_extension_test.py`:
```python
import some_extension
# ...
```

### DIFF
--- a/waflib/extras/pytest.py
+++ b/waflib/extras/pytest.py
@@ -136,11 +136,13 @@ def pytest_process_use(self):
 			if not isinstance(tg.link_task, ccroot.stlink_task):
 				extend_unique(self.pytest_dep_nodes, tg.link_task.outputs)
 				extend_unique(self.pytest_libpaths, tg.link_task.env.LIBPATH)
-				extend_unique(self.pytest_libpaths, [tg.link_task.outputs[0].get_bld().parent.abspath()])
+				bldpath = tg.link_task.outputs[0].get_bld().parent.abspath()
+				extend_unique(self.pytest_libpaths, [bldpath])
 
 				if 'pyext' in tg.features:
 					# If the taskgen is extending Python we also want to add it o the PYTHONPATH.
 					extend_unique(self.pytest_libpaths, tg.link_task.env.LIBPATH_PYEXT)
+					extend_unique(self.pytest_paths, [bldpath])
 
 
 @TaskGen.feature('pytest')


waflib/extras/pytest.py
```diff
@@ -136,11 +136,13 @@ def extend_unique(lst, varlst):
 			if not isinstance(tg.link_task, ccroot.stlink_task):
 				extend_unique(self.pytest_dep_nodes, tg.link_task.outputs)
 				extend_unique(self.pytest_libpaths, tg.link_task.env.LIBPATH)
-				extend_unique(self.pytest_libpaths, [tg.link_task.outputs[0].get_bld().parent.abspath()])
+				bldpath = tg.link_task.outputs[0].get_bld().parent.abspath()
+				extend_unique(self.pytest_libpaths, [bldpath])
 
 				if 'pyext' in tg.features:
 					# If the taskgen is extending Python we also want to add it o the PYTHONPATH.
 					extend_unique(self.pytest_libpaths, tg.link_task.env.LIBPATH_PYEXT)
```
kljohann: When making this change I noticed this line and the corresponding comment. Is this a typo? Should it also be `pytest_paths` here? And what exactly is `LIBPATH_PYEXT`? I only found two references in this repository and its purpose is not clear to me yet.

xbreak: Yeah, that's a typo. The `LIBPATH_PYEXT` contains the Python interpreter libpath (needed for Python extensions).

ita1024: @fedepell FYI ^

ita1024:
Files are not always written to the build directory, it is really necessary to have `tg.link_task.outputs[0].get_bld().parent.abspath()` and not just `tg.link_task.outputs[0].parent.abspath()`?

xbreak:
>Files are not always written to the build directory, it is really necessary to have tg.link_task.outputs[0].get_bld().parent.abspath() and not just tg.link_task.outputs[0].parent.abspath()?

You're right, just the node `parent.abspath` should do it.

Additionally, I think it's more correct to only add that path to `pytest_libpaths/LD_LIBRARY_PATH` if it is *not* a Python ext. Because if it's an extension in needs to go in `pytest_paths/PYTHONPATH` instead. However, as I noted above we cannot really know at this point (there's no `install_from` attribute that tells us where the root is), which is the correct path to use as it could be nested inside a package or not.

waflib/extras/pytest.py
kljohann: Thanks for the pointer. Using waf 2.0.3 and  adding `pytest_path=[bld.path.get_bld()]` to the `some_extension` target in the example above this patch is no longer necessary. Thanks for documenting those arguments in f336f0bf.

kljohann:
I'm closing this PR as this use-case can be solved by following the instructions in f336f0bf (which is included in waf 2.0.3).

PR #2072, opened 2017-11-08
fix load_tool error messages and cover all occourences

fedepell:
When a tool cannot be loaded the wrong path was displayed on the error
message. sys.path was always displayed but the actual path used depends
on tooldir being passed and on the value of with_sys_path parameter.

I put the exception handling (raising the fatal) inside load_tool itself
as this is the only place where the exact path is known, without having
to recalculate it outside. To be able to use fatal there also the ctx
has to be passed from the various points.

In this way all load_tool exceptions are caught and reported, while before
an exception during load from configure or load_special was not caught for 
example (and as far as I see was not handled later too, so would just result in 
a Python traceback), just  during the options. I don't know if this is not desirable 
we could  just up the exception with additional path information and leave the 
exception handling at the level it was before.


ita1024:
The `ctx` parameter is meant to be removed and is only present for compatibility reasons. There are valid reasons for loading modules without the presence of context objects.

The `ctx` object is an optional parameter of `load_tool` function but it becomes mandatory with these changes, so this cannot applied at this point (API change). Furthermore the ImportError exception is no longer raised (API change). This would affect `Context.load` and `Context.load_special_tools` too.

fedepell:
Thanks for the explanation, all clear and I was also worried the API change would be a problem.

I updated the branch and give it another try: I just catch ImportError internally ini load_tool, add an attribute with the sys.path effectively used, and rethrow it. API and behaviour wise should be the same, but externally the new attribute can be used to notify correctly the user of the path used (and I added the printing of this in the Configure) and not just a generic sys.path.

Hope it may be better now.

Thanks!
F.

PR #2076, opened 2017-11-13
Fix recursion bug in TaskGen.__repr__ if task_gen field references self

justinfx:
This is a merge request to fix a hard-to-debug issue that one of our users managed to trigger. It happens if a TaskGen is displayed in the console and the instance happens to somehow have a field that refers to itself. 

```
task = bld(...)
task.x = task
```

The result is an infinite recursion (until Python interp kills it). I realize that this is a very uncommon thing to do intentionally. But it turned out the user had a typo in their wscript that managed to produce this result.

### DIFF
--- a/waflib/TaskGen.py
+++ b/waflib/TaskGen.py
@@ -94,7 +94,7 @@ class task_gen(object):
 		"""Debugging helper"""
 		lst = []
 		for x in self.__dict__:
-			if x not in ('env', 'bld', 'compiled_tasks', 'tasks'):
+			if x not in ('env', 'bld', 'compiled_tasks', 'tasks') and self.__dict__[x] is not self:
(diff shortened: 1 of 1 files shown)


ita1024:
How effective is this change at preventing recursion? The outputs would not show the field(s) that are causing problems, would they?

justinfx:
I agree that it isn't comprehensive. It doesn't solve the case where any of the field values could also have a reference back to the task gen. I would say that it is effective at preventing one cause of a recursion bug.

ita1024:
I agree with the overall intention, but the changes should not be added as-is: the attribute no longer appears in the outputs so this conceals the problem rather than reveals it. Besides that, recursion errors can still occur if objects are nested so the fix is ineffective.

justinfx:
Got it. Thanks for reviewing.

PR #2080, opened 2017-11-26
Fix building shared libraries with -fPIE

rlaager:
I'm seeing this with NTPsec, which uses waf to build.

Debian wants to build with -fPIE for security reasons.  This is not compatible with building shared libraries like the ntpc python extension.  If the -fPIE flag comes before -shared (or -fPIC), the latter will override, and the build succeeds.


rlaager:
NTPsec is reporting these flags:
  LDFLAGS                       :  -Wl,-Bsymbolic-functions -fPIE -pie -Wl,-z,relro -Wl,-z,now -Wl,--as-needed -lbsd -lssp_nonshared -Wl,-z,now -Wl,-z,relro 
  LINKFLAGS_NTPD                :  -pie

ita1024:
Such a change cannot be applied because it would break a lot of other projects. Could you add your flags to LINKFLAGS instead, for example `LINKFLAGS=-fpie waf configure build`?

rlaager:
LDFLAGS is a pretty standard name for such things. But yes, it seems to work with:
`LINKFLAGS='$(LDFLAGS)' LDFLAGS='' python waf configure ...`

ita1024:
The matter seems to be answered, closing this now.

PR #2084, opened 2017-11-29
javaw: clean up CLASSPATH generation with dependencies

fedepell:
when building up the CLASSPATH in some cases empty strings were added or
repeated paths, leading to a quite ugly and long CLASSPATH. this patch should
not append empty or duplicate entries.

from the java demo we had before generated for example the following two:
.:..:aaaa.jar:bbbb.jar:.:..:
aaaa.jar:bbbb.jar::

While now we have the more correct:
.:..:aaaa.jar:bbbb.jar:
aaaa.jar:bbbb.jar:


ita1024:
- The change may introduce errors of the form `'NoneType' object is not iterable`
- There is a method called `env.append_unique`

fedepell:
Thanks for the feedback! Hopefully better now, I fixed the commit!

Used append_unique and made sure it works also if CLASSPATH is a string as "aaaa.jar bbbb.jar" which is also valid.

Thanks!

ita1024:
Splitting would make sense if we were certain that existing paths never contained spaces. Cannot we just use the changes from the master branch?

fedepell:
Thanks for the changes on the master. For the double entries is very fine as you did. But without the check for empty strings I still get:

`07:28:15 runner ['/usr/bin/javac', '-classpath', 'aaaa.jar:bbbb.jar::', '-d', '/home/fede/waf/fede-waf/waf/demos/java/build/animals', '-Xlint:unchecked', '-source', '1.7', '-sourcepath', 'animals', '../animals/src/org/example/Animal.java']`

I corrected my branch (and rebased to master) for this if you want to apply, so you get:

`07:32:21 runner ['/usr/bin/javac', '-classpath', 'aaaa.jar:bbbb.jar:', '-d', '/home/fede/waf/fede-waf/waf/demos/java/build/animals', '-Xlint:unchecked', '-source', '1.7', '-sourcepath', 'animals', '../animals/src/org/example/Animal.java']`

Just to give some context: I am using the CLASSPATH to build a series of parameters to a coverage library (JaCoCo) and this empty value is annoying it, so I have to filter it afterwards. But I guess filtering at the beginning would be better so whoever uses it later has it correct. Thanks!

PR #2085, opened 2017-11-29
buildcopy: Create destination directories as needed

xbreak:


ita1024:
Folders are created automatically when using `Node.find_or_declare`, so what are you doing exactly?

xbreak:
I'm copying exiting source nodes to the equivalent build nodes. Currently the equivalent build node is created with "`<srcnode>.get_bld()`". Should this be replaced with `find_or_declare` you mean?

ita1024:
I mis-read the change. This should do.

PR #2086, opened 2017-12-01
cxx: Added C++ headers .hpp, .hxx and .hh to cxx.ext_in.

xbreak:
This PR aligns the `cxx.ext_in` to use the same list of header extensions as in `qt5` and `TaskGen.HEADER_EXTS` (`['.h', '.hpp', '.hxx', '.hh']`).

To provide some motivation, this was traced down to be the problem in a scenario with `waflib.Errors.TaskNotReady` errors when a C++ source file included a generated header (with extension `.hpp`). The same error was reproduced using the `protoc` feature, modified to produce `.hpp` files instead of `.h`.


ita1024:
Which classes are you referring to?

`ext_in` is a symbol and is not a calculated value, so headers-producing tasks are expected to specify `ext_out=['.h']` exactly like in https://github.com/waf-project/waf/blob/master/demos/c/program/wscript_build#L5

xbreak:
Regarding `protoc`? If yes, https://github.com/waf-project/waf/blob/master/waflib/extras/protoc.py#L84
Otherwise I'm referring to an internal tool that works very similar to `protoc` with an `ext_out=['.hpp']`. The ordering by file extensions in `Task.is_before` https://github.com/waf-project/waf/blob/master/waflib/Task.py#L852 will not work in this case because the `cxx` task does not list `.hpp` as an `ext_in`. Or am I misunderstanding this?

Alternatively, are you saying that a task with a `*.hpp` target should specify `.h` in `ext_out`? I.e, it has less to do with actual extensions and more to do with naming conventions?

ita1024:
This is indeed a naming convention, that task with a `.hpp` target should specify `ext_out=['.h']`. There are just too many header extensions to have a complete list ('.H' is also missing for instance), and this affects performance.

xbreak:
Ok, knowing this, the following paragraph in the waf book takes  on a clear meaning:

> The extensions ext_in and ext_out do not mean that the tasks have to produce files with such extensions, but are mere symbols for use as precedence constraints.

Maybe this could be clarified even more by using a real example of an "in-use convention".

Cheers

PR #2091, opened 2017-12-13
Added task target to output of parallel_debug.

przemekr:
Without the task name/target it is not obvious which steps in the build take
the most of the build time.

Tested the SVG output with Firefox 52.


ita1024:
Cool. It seems that some spaces were inserted instead of tabs, so the changes may be amended.

przemekr:
oh, sorry, I should pay more attention to white spaces.

ita1024:
No worries. It was in the javascript part anyway; in Python the tabs/space mix tends to create problems.

PR #2092, opened 2017-12-14
Fix msvc support for qt5.py

mclayton7:
Clear unsupported flags for qt5 feature check.

Visual Studio doesn't support the ```-std=c++11, -fPIE, -fPIC``` flags, so those should be removed for the feature check to compile.


ita1024:
If the first test builds, then the other compilation tests are not executed, are they?

mclayton7:
You're right, sorry!

PR #2095, opened 2017-12-18
Udated Erlang support.

przemekr:
This adds a scanner method to track Erlang headers dependencies.
Support for EUnit tests
Support for EDocs
Support for ERL, ERLC, ERLC_FLAGS environment settings.


ita1024:
It looks promising, but there are a few questions:

1. What would happen if self.app is not defined in `'${PREFIX}/'+getattr(self, 'app', False)+'/ebin'`? Is an error message missing somewhere?
2. Is `import threading` really necessary?
3. What are the reasons for expanding `erl.run_str` as `erl.run`? Setting `task.cwd` is possible.
4. Is the `print ` line intentional? This prevents this module from being used with Python3.
5. Are you aware that tabs and spaces are mixed in the resulting file?
6. The flags `'-boot', 'start_clean', '-noshell', '-noinput'` look hard-coded
7. `erl_test` tasks will not be executed again when the contents of `test_modules` change. If this is not intentional, try using `run_str = "${TEST_MODULES}...` and setting the corresponding `task.env.TEST_MODULES` instead

Can you consider the above and perhaps update this pull request? Thanks.

przemekr:
Hi, thank you for the review.  Yes, I will update the pull request. 
1) should be removed, it is part of functionality I had in another project with support to install erlang 'applications'
2) its a leftover from the Erlang ANS.1 compiler which wasn't thread multi process safe (but here we don't include the ANS.1 compiler support so threading should be removed)
3) agree, I will update
4) yes, I will remove
5) I tried to use tabs, only in the string passed to erlang command to evaluate/run the test code there are spaces (but should be ok to change to all \t)
6) yes, but that is needed to run erlang in non-interactive mode and get the proper exit code. 
7) as it is not easy for waf to guess what are the tested modules I added the 'modules under test' (ie. hello.erl) as sources with the '.beam' extension. (so the test run is dependent on the hello.beam, which is depended on hello.erl)

przemekr:
Oh, sorry, not sure why your commits got included in the pull request. I thought It would be good to rebase on top of origin/master.

ita1024:
There is a conflict caused by the extra commits included in this pull request. Would you be so kind so as to actually rebase your commit on top of the updated version of origin/master?

Note: point 6. will lead to a rewrite because such flags should be configurable and never be hard-coded in the middle of the file. The answer to point 7. does not justify the expansion of the run_str attributes into complex hand-written methods so those shall be rewritten as well.

przemekr:
I amended/rebase the commit on top of the original parent. 

regarding 6) I tried to clean up the flags and move it to separate place, it looks better for me. I don't think this should be entirely user configurable (as, lets say '-c' flag is also 'hard-coded' in gcc or cc_root).

regarding 7) Most likely I don't know how to do it in a nice way with run_str. (sorry still learning...)  I like the 'def run()' method as it allows to break the task command creation to several steps. But of course please update!

Issue #2098, opened 2018-01-13
Cpu extension detection


bjourne:
It would be useful if the environment detection done by the C tool included cpu extensions. For example, on my machine:

    print (ctx.env.DEST_OS, ctx.env.DEST_CPU)
    ('linux', 'x86_64')
    print ctx.env.DEST_CPU_EXTS
    set(['see', 'avx2', 'avx', 'sse2', ...])

If you don't mind, I'm gonna make a pull request for it.

ita1024:
Such detection is best done with configuration tests. Support in waf cannot support all compilers and is likely to be unusable except for a few cases.

bjourne:
I tried that first, but I had to write one configuration check for each extension. Like this

    ctx.check(msg = 'Checking for AVX support',
              execute = True,
              fragment='''int main() {
              #ifdef __AVX__
              return 0;
              #else
              return 1;
              #endif
              }\n''', ...)

It becomes tedious if you have a whole bunch of cpu extensions you want to check for. Perhaps there is a better way I don't know about?

ita1024:
Have you tried using a loop over the required CPU extensions?

ita1024:
We will not apply such changes because more accurate configuration tests can be written rather easily.

PR #2099, opened 2018-01-13
Support for DEST_CPU_EXTS #2098

bjourne:
Here is the patch! I have only added those extensions which I think are most useful to detect; popcnt, see and the avx families. More can of course be added. To use the features in compilation one of course has to use the `-march=native` option.

### DIFF
--- a/waflib/Tools/c_config.py
+++ b/waflib/Tools/c_config.py
@@ -70,6 +70,18 @@ MACRO_TO_DEST_CPU = {
 '__xtensa__'  : 'xtensa',
 }
 
+MACRO_TO_DEST_CPU_EXTS = {
+        '__ADX__' : 'adx',
+        '__AVX__' : 'avx',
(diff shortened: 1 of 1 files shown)


bjourne:
1. Yeah. I had to add it because running `ctx.env.append_unique('CFLAGS', '-march=native')` and then `ctx.load('compiler_c')` didn't work. That could be a bug, I think. But I can remove that.
2. These particular macros should all work on gcc/clang/icc. msvc is different, as always.
3. There are a handful more one can add, but those I added are the ones people care about.
4. It is based on the same mechanism as `DEST_CPU` so `DEST_CPU_EXTS` would only be misleading if `DEST_CPU` is.

ita1024:
Concerning point 4., current `DEST*` variables are for the target system, yet the detected flags are for the host system (`-march=native`).

ita1024:
The problems with this pull request are not addressed so it will not be merged.

PR #2107, opened 2018-02-12
Added opt decorator to make OptionsContext extendable (using @opt).

daniel-brosche:
Extended OperationContext in a similar way as ConfigurationContext to increase extensibility. Over this way it is possible to write tools that adds new functions to the OperationContext class (via @opt decoration).


ita1024:
Do you have any concrete example? Options do not have any `mandatory` parameter as far as I know.

daniel-brosche:
In my case I need some general helper functions (provided by a loadable tool) for the `OperationContext` that are currently only available for `ConfigurationContext` and `BuildContext` class.

```python
def configure(conf):   
    if conf.is_toplevel():
          pass
```

To make the new `is_toplevel` function accessible in the `OperationContext` I need a similar mechanism as in the @conf-way, like:

```python
from waflib.Configure import conf
from waflib.Options import opt

@opt
@conf
def is_toplevel(self):
    return len(self.recurse_cache) == 1 #self.srcnode == self.path

```

ita1024:
The implementation does not really work because it specifies `mandatory` which is only for configuration tests. The decorator is not options-specific at all, and a direct binding would make more sense:

```
from waflib.Context import Context
def is_toplevel(self):
    ...
Context.is_toplevel = is_toplevel
```

Adding an annotation for all context methods could make sense, but this rarely occurs in practice. One decorator per context class is not going to work in the long run either.

The changes introduce invalid code (`mandatory` tests copy-paste), introduces changes on the wrong level (`OptionsContext` versus `Context`), and does not provide any sensible simplification at this point so they will not be merged.

daniel-brosche:
Nearly in all tools the decorator `@conf` has been used to introduce new functions to `ConfigurationContext` and `BuildContext`. Besides the fact that mandatory tests does not make sense for `OperationContext` my idea was to introduce an analog options-specific way to extend it. I see some benefits for an author to have analogous mechanism (code patterns) to extend the different context classes. 

Another thing were I was struggeling was that `@conf` extends the `BuildContext` class that implied for me that the base for this class is `ConfigurationContext` which isn't the case. So relation between this two classes is for me not obvious and it was for me not easy to understand why I'm not able to extend the `OptionsContext` in the same way.
I asked myself why not to introduce an `@build` annotation to extend explicitly the `BuildContext`, too?
Over this way an tool-author has also the possiblity (fine grained control) to just extend `BuildContext` class with functions that may access buildcontext-specific attributes.

Make this sense or do I understand not whole context?
Has a modified decorator without mandatory tests a chance to be accepted?

daniel-brosche:
My idea was something like:

```python
def ctx(f):
	setattr(Context.Context, f.__name__,f)
	return f

def bld(f):
	setattr(Build.BuildContext, f.__name__,f)
	return f

def opt(f):
	setattr(Options.OptionsContext, f.__name__,f)
	return f
```

ita1024:
1. The names `ctx`, `bld` and `opt` are widely used for context instances so the names will clash with most scripts
2. Method can be bound directly without a specific decorator per context class
3. There are many other context classes so for consistency, so by the same reasoning there should be one decorator per Context class `@dist`, `@distclean` which is of course unnecessary
4. Most `@conf` configuration tests can be called in the build phase and it is usually desirable to silence certain configuration tests selectively. Binding methods to context classes is a very different matter.
5. There is no visible need for such decorators

The net outcome is that the changes that you suggest bring more code to maintain without any actual use in any part of the framework or in the example. Such changes cannot be merged.

daniel-brosche:
That view that I explained is from a fresh perspective working with waf from time to time since some months. There are many concepts not clear in depth and it is hard to write good tools following some patterns. The mentioned decorators (in general despite the naming) seems a good thing to establish patterns where poeple can align without to much exceptions. Thx anyway...

PR #2110, opened 2018-02-16
Adding ability to specfy if install task should log output on successful install.

nstemmler:
Simple change that adds the ability to specify if an install task should log output, with the default option being true to match current behavior. I named the parameter 'log_success' to try to indicate this is only for suppressing successful installation log output; let me know if you will accept this change but would like me to rename the parameter first.


ita1024:
Interesting idea. There are a few concerns though:
1. The name "success" is terribly misleading
2. Symlink installation is ignored
3. There are no tests to reduce the possibility of regressions

Since the methods can be overridden in your own context subclass, this is not urgent. Adding a method to log installation operations may be a better design than such a variable.

PR #2115, opened 2018-03-01
protoc: handle case of proto files in deep directories when output is generated with paths

fedepell:
When the proto file is in a directory structure the generated header and source are also generated in a similarly named structure and therefore they will not be later found.
A complete description of the behaviour can be found here:
https://github.com/google/protobuf/issues/1743
The patch adds a -I to the full directory containing the proto file as the first parameter to address the issue as explained in the link. This cannot be solved by just passing it in the includes as it doesn't work for multiple .proto files in different directories in the same bld (ie including "a/b/file1.proto c/d/file2.proto" as first needs -Ia/b second -Ic/d).
I added a simple example in the playground to test the behaviour and compare with old one.


ita1024:
It would be nice to use `${PROTOC_ST}`, but perhaps all protobuf-like compilers only supports `-I` on all platforms?

fedepell:
I changed as you suggested and put a small comment about it. 

I tested on a couple of cases and works fine!

ita1024:
Nice!

PR #2121, opened 2018-03-07
xcode6: fix py3 support

andrewkeller:
Avoids exception in the `xcode6` extra when Waf is ran on Python 3.


andrewkeller:
@ita1024 This patch appears to be relevant to Waf 1.8.17 and later.  If you'd like this patch rebased against and older branch, or any of the maintenance branches, feel free to let me know.

PR #1716, opened 2016-03-11
Correction of boost module detection of Boost.Log

cawka:
The previous logic in #1709 made an incorrect assumption that the filename of shared/static library indicates that it was build as multi-threaded or single threaded.  This assumption does not hold in many Linux distributions.

In addition to that. Boost.Thread and Boost.Log require -pthread (or some other) flags in order to properly link.

PR #2123, opened 2018-03-08
Correction of boost module detection of Boost.Log

cawka:
The previous logic in #1709 made an incorrect assumption that the
filename of shared/static library indicates that it was build as
multi-threaded or single threaded. This assumption does not hold in many
Linux distributions.

In addition to that. Boost.Thread and Boost.Log require -pthread (or
some other) flags in order to properly link.


cawka:
This is an updated version of https://github.com/waf-project/waf/pull/1716

ita1024:
`self.end_msg('None')` on line 382 is not meant to be called, is it? Would it make sense to raise an exception instead, such as `self.fatal('No pthread support')`?

cawka:
Hmm. Probably, but not sure.  I was thinking of the case when no flags required, but that already handled by the empty flag...

PR #2126, opened 2018-03-11
Use `return` instead of `raise StopIteration`

mote-s:
Python 3.7 [changes the behavior of `raise StopIteration` in generators](https://docs.python.org/3.7/whatsnew/3.7.html#changes-in-python-behavior).


ita1024:
Is it not possible to remove those lines entirely? (empty `return` statements)

PR #2133, opened 2018-03-21
Implement '__hash__' and '__eq__' methods on Node to make it dict-friendlier

captainmalloc:
Hi,

When node instances are used as key in a python dictionary the runtime environment is not always capable of retrieving them (I am using Python 2.7.8).

This has 2 consequences. The first is the less problematic one: in custom dictionary structures it is safer to use 'aNode.abspath()' to correctly index the dictionary and retrieve the corresponding node with 'Node.find_node()'. 
The second consequences is quite more subtle. The 'waflib.Build.store()' stores the 'node_sigs' dictionary in the Pickle DB and the 'waflib.Build.restore()' loads the pickle DB at the end of a Waf run in the 'node_sigs' on a successive builds. The python runtime environment is not always able to recognize the loaded nodes in the restored version of the 'node_sigs' structure thus causing the 'Task.runnable_status()' to return a 'RUN_ME' code because 'must run: an output node has no signature'.

This simple pull request implements the '__hash__()' and '__eq__()' methods in the 'waflib/Node.py' class to make it dictionary-friendlier.

### DIFF
--- a/waflib/Node.py
+++ b/waflib/Node.py
@@ -168,6 +168,12 @@ class Node(object):
 		"""
 		raise Errors.WafError('nodes are not supposed to be copied')
 
+	def __hash__(self):
+		return hash((self.name, self.parent))
+
+	def __eq__(self, other):
+		return (self.name, self.parent) == (other.name, other.parent)
+
 	def read(self, flags='r', encoding='latin-1'):
 		"""
 		Reads and returns the contents of the file represented by this node, see :py:func:`waflib.Utils.readf`::


ita1024:
Can you please provide a testcase to illustrate the problem?

ita1024:
Note: Such methods were implemented before and have a significant performance impact. They are also only needed if Node objects are mis-managed, such as in cross-context data sharing or questionable persistence approaches. The change suggested will not be applied.

captainmalloc:
Ok. I actually created a small example where the same node is 'make' twice on 2 different task_gens  and used as index in a dictionary (see below). Everything works fine.

I guess you are right, I have been fooled by cross-context data sharing and/or wrongly-managed parallel node creation.

Sorry about the push-request.

```
#!/usr/bin/env python
# encoding: utf-8

import waflib

PRODUCED_FILE_NAME = 'a_produced_file_%s'
CONSUMED_FILE_NAME = 'a_consumed_file_%s'

def configure(cnf):
    cnf.env.MAX_SEED = 1000
    cnf.env.NODE_CACHE = dict()

@waflib.Configure.conf
def buildNode(cnf, seed):
    aNode = cnf.bldnode.make_node(PRODUCED_FILE_NAME % seed)
    if not aNode in cnf.env.NODE_CACHE:
        cnf.env.NODE_CACHE[aNode] = 1
    else:
        cnf.env.NODE_CACHE[aNode] += 1
    return aNode

def build(bld):
    if not bld.targets:
        bld.targets = '*'

    bld(features='producer', name='producer')
    bld(features='consumer', name='consumer')
    def cacheDumper(bld):
        print '--- This section should be empty ---'
        for key,value in bld.env.NODE_CACHE.iteritems():
            if value == 1:
                print '%s : %d' % (key, value)
        print '--- End section ---'
    bld.add_post_fun(cacheDumper)

@waflib.TaskGen.feature('producer')
def process_producer(self):
    for i in range(self.env.MAX_SEED):
        tgtNode = self.bld.buildNode(i)
        self.create_task(
                'produce',
                tgt=[tgtNode]
        )

@waflib.TaskGen.feature('consumer')
def process_consumer(self):
    for i in range(self.env.MAX_SEED):
        inNode = self.bld.buildNode(i)
        outNode = self.bld.bldnode.make_node(CONSUMED_FILE_NAME % i)
        self.create_task(
                'consume',
                src=[inNode],
                tgt=[outNode]
        )

class produce(waflib.Task.Task):
    def run(self):
        return self.exec_command('echo "this is a produced file" > %s' % self.outputs[0].abspath())

class consume(waflib.Task.Task):
    def run(self):
        return self.exec_command('echo "this is a consumed file" > %s' % self.outputs[0].abspath())
```

ita1024:
1. `cnf.env.NODE_CACHE` will lose data because the whole Node tree is not saved during serialization.
2. `Node` objects cannot be shared across contexts. Each context has a distinct Node subclass, and those are incompatible.
We will try to make this more clear in the documentation as a remediation.

captainmalloc:
Hi,

Thanks for your explanation, but let me just spend few words on my example and this pull-request, just for the record.

The point of the example was just to demonstrate that Node instances can be safely used as keys in a dictionary. I didn't expected `cnf.env.NODE_CACHE` to be serialiazed with data. 
In the example the `buildNode()` method is called always twice for the same filesystem path, by the defined TaskGen callbacks `process_producer()` and `process_consumer()`. The first time a record will be created in the `cnf.env.NODE_CACHE`, the second time the same record is supposed to be used and the counter incremented.
If the bug I was looking for was there, I would have been able to build two Node instances pointing to the same filesystem path having their associated counter set to 1 in `cnf.env.NODE_CACHE`. This would have been shown by the `cacheDumper()` `post_fun` at the end of the flow.

The example is related to this pull-request, because I got in a similar situation with the `waflib.Build.node_sigs`, a dictionary that uses Node instances as keys.
At the end of the flow execution, the `waflib.Build.node_sigs` correctly contained all the Nodes generated by the Tasks in my flow. The `waflib.Build.store()` was correctly serializing the `node_sigs` and, on successive rebuild, the `waflib.Build.restore()` was correctly deserializing it. But during the rebuild the `Task.runnable_status()` was not not able to find the expected Node in the dictionary (even if a Node instance containing the expected path was there) and the task generating it was being rebuilt.
During my debug session, that lead to this pull-request, I put a breakpoint at the end of `waflib.Build.store()` and manually restored the serialised data as `waflib.Build.restore()` would do. I actually verified that a particular node was there in the serialised data and I wasn't able to retrieve it in the temporary de-serialized one (even if a Node pointing to the expected path was there in the de-serialized dictionary).
When I applied the proposed patch, the  behaviour was corrected and `node_sigs` behaviour was more coherent.

In conclusion. I wasn't able to create a simple test-case showing my point. But I still think there is something that went wrong in my flow execution and that should be prevented by the framework.

Thanks for your time

PR #2134, opened 2018-03-21
gdbus: Add glib DBus bindings

JPEWdev:
Adds support for generating dbus bindings using the newer gdbus-codegen
tool that is designed to replace the deprecated dbus-binding-tool tool


ita1024:
It seems a bit early to remove the existing tool, however obsolete it is called. The code will be moved to extras/playground.

PR #2135, opened 2018-03-22
protoc: add .proto dependencies also in include path otherwise protoc will fail if dependency is in another directory

fedepell:
protoc extra already correctly had the scanner for dependencies and would build dependent .proto files. But if the dependency is in another directory then also the include path to that directory has to be passed to protoc via -I otherwise the file will not be found and execution will fail.

Added also a small example in protoc playground.


ita1024:
Prefer `abspath` to `bldpath` because the task execution directory not match the build directory (`Task.get_cwd()`).

PR #2146, opened 2018-04-06
Detect architecture-specific library path for Boost on dpkg-based systems

eric135:


ita1024:
1. `subprocess.check_output` is not present in Python 2.5-2.6.
2.  there are encoding concerns in Python 3

Could you try `conf.cmd_and_log(['ls'])`? I could update the changes but I have no system to test the resulting code. Thank you!

eric135:
@ita1024 `self.cmd_and_log(['ls'])` saves the output of ls to build/config.log

eric135:
I modified the commit and now it should work on Python 2.5-2.6 (my system runs 2.7 so I wasn't able to test it).

ita1024:
1. Have you tried on Python3?
2. The point is precisely to leave a trace in `config.log`, but this can be disabled if deemed necessary. What would be the reason to not have it?

eric135:
I added the `config.log` trace and modified the latest version to also work under Python3. I verified on Python 3.6.3.

ita1024:
Apologies for the confusion! A single line is all that is needed!
```python
deb_host_multiarch = self.cmd_and_log([self.env.DPKG_ARCHITECTURE[0], '-qDEB_HOST_MULTIARCH'])
```

ita1024:
Please update your copy and try https://github.com/waf-project/waf/commit/bef6f1b4bfe5527cbfad550b46bedb7e2a3c1352

eric135:
I copied the changes from bef6f1b into one of my projects using waf and it detected the architecture-specific Boost install using both Python2 and Python3.

ita1024:
Thank you!

PR #2148, opened 2018-04-12
Better support for cross compilation with clang

DragoonX6:
These changes should make it easier to cross compile a project without editing the wscript to support the particular target the user is trying to compile for.

It also implements support for the MSVC mode, allowing one to compile MSVC compatible binaries, or compile a library using the MSVC ABI.  
It's a bit more of a manual process using it, but I think assuming more is a bad idea, since for example MinGW's CRT could be used, and making more assumptions would make that harder to do.

Please review, and let me know if you need anything to be changed.


ita1024:
1. Ok for the introducing the `def_PATTERN` variable.
2. C and C++ support are meant to be independent from each other, but a dependency from `clangxx` is introduced to `clang.py` and c++-related variables are introduced in `clang.py`
3. `conf.env.CXX += ['-target', conf.options.clangxx_target_triple]` is better written as `conf.env.append_value('CXX', ['-target', conf.options.clangxx_target_triple])`
4. The changes to `clang.py/clangxx.py` will not be applied because we will basically need to support the options forever and there is no documentation, environment setup, examples or tests to help with that new maintenance work. Can you move your clang code to your own extension? (there are examples such as `waflib/extras/c_emscripten.py`). If not, then I might do that.

DragoonX6:
I can easily change 2 and 3, but would it be okay to live in `clang.py/clangxx.py` if I write documentation, examples, and tests?  
I don't know what you mean with environment setup.

ita1024:
The `def:` string change is applied - see f8c86e0895783f0c784020f37f1589adc079cc95

No changes will be applied to `clang.py/clangxx.py` at this point, so the other changes must go to a specific extension (under `waflib/extras/`).

DragoonX6:
Alright, I'll push an update soon.

PR #2150, opened 2018-04-15
eclipse: handle case where python install_from is a Node object

fedepell:


ita1024:
Does that work if the `Node` object points to an arbitrary folder somewhere else on the filesystem? (nested folders).

PR #2157, opened 2018-04-28
can't create the output folder when path was unicode

messiLiao:


ita1024:
The `unicode` type only exists in Python2, so this change would break Waf for Python3. There are also many more areas where a similar change might be required.

Why is such change deemed necessary? What are you trying to achieve?

messiLiao:
Oh, I ignore the python3. Yeah, you are right, I can't fix the issue in this way. My path was all ascii characters. but the variable lst was unicode. Let me dig it out, and try another way.

ita1024:
We will not be applying such change. If proper unicode support is required, use Python3 or convert to the str type.

PR #2162, opened 2018-05-09
eclipse: add generation of external jar dependencies for java projects to eclipse file

fedepell:


fedepell:
Just as a minor note: the playground eclipse example already imports some external JARs so that example already demonstrates/covers this addition.

PR #2164, opened 2018-05-11
eclipse: generate an external builder when no CDT is used in the project

fedepell:
When CDT is not included in the project (ie. we just have Python and/or Java) the current implementation would not create automatically a call to waf for the build stage. This patch adds in such cases an external builder that automates the call to waf without the need to manually configure one.

### DIFF
--- a/playground/eclipse/wscript
+++ b/playground/eclipse/wscript
@@ -23,7 +23,12 @@ This will generate the needed configuration files for Eclipse:
 -) .project is generic Eclipse project file
 -) .cproject for C/C++ CDT
 -) .classpath for Java JDT
--) .pydevproject for Pydev)
+-) .pydevproject for Pydev
+
+If CDT is in the project (so at least one C/C++ build) then CDT builder
+will be used to call waf as it is more versatile. If just JDT/PYydev are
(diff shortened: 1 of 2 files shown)


ita1024:
1. Calling a parameter "waf" is ambiguous, what is it for?
2. The method `impl_create_project` has 6 parameters after this change. What should be done to avoid adding a 7th parameter?
3. Have you considered `%` instead of `+` concatenation in `'<project>/' + extbuilder_dir + '/'+extbuilder_name`?

fedepell:
Thanks for the feedback!

1. I agree but I kept it same as the already existing self.impl_create_cproject. The parameter actually refers to the waf executable (in both cases, which is os.path.abspath(sys.argv[0]))
2. I can just use os.path.abspath(sys.argv[0]) inside the function, just kept it like this to make it same to the already existing function
3. I can fix that, no problem, will commit shortly!

fedepell:
Fixed 3. For 1/2: I can maybe rename both cases with something like "waf_exec" or just remove and use directly the os.path.abspath(sys.argv[0]).

ita1024:
A better name is really needed. I can apply this now and let you think of a better name in the future.

PR #2165, opened 2018-05-11
eclipse: give a better name to waf variable indicating the waf executable path

fedepell:
See also comments in #2164


ita1024:
Thank you!

PR #2166, opened 2018-05-11
eclipse: make all generated file utf-8 encoded

fedepell:
eclipse generates them by default in utf-8 (or will resave them
as such the moment they are touched) so this is more conformant
to eclipse format and so we can support also utf-8 project names
and such from waf

implementative note: I have to pass flags explicitly as "wb" otherwise
I get the failure when in Utils writef:

```
        if sys.hexversion > 0x3000000 and not 'b' in m:
                data = data.encode(encoding)
                m += 'b'
        with open(fname, m) as f:
                f.write(data)

```
The data.encode is called and since data is bytes this method is not
present. I don't know if we should put a check there and avoid calling
encode in such cases (or catching the exeception).


ita1024:
1. Have you tried this in Python 2?
2. Have you considered using a helper function instead of copy-pasting?

fedepell:
Hi,
1. Yes I tried on a 2.7.6 and it works fine. If you have doubts for earlier I can try to check (from the docs it should be fine)
2. Let me check right now!

fedepell:
Cleaned a bit up number 2 with a helper function (not more compact, but for sure more readable)

Hope it's fine!

PR #2176, opened 2018-06-08
Handle integers in get_flat of ConfigSet

danbar:
Handle integers in method `get_flat` of module `ConfigSet`.
Otherwise, a `TypeError: can only join an iterable` would be raised if the configuration set contains integer values.


danbar:
Was this pull request closed due to the recent move to GitLab?
If yes, should I move the pull request to GitLab?

ita1024:
All pull requests have been closed due to the move to GitLab.

Regarding this particular pull request, the changes create an inconsistency: tasks assume that only strings or list of strings are handled. This affects performance as well, and opens a dangerous precedent as other types (float, unicode, bytes) would also need to be handled.

For these reasons I suggest that the method `get_flat` should be overridden in your project deemed necessary.

danbar:
Ok, thx for the clarification.

